PR #5, opened 2017-09-07
Split view and test files to group related endpoints

allait:
Splits views/auth into:
* views/auth - login/logout endpoints
* views/create_user - endpoints for completing user registration
* views/reset_password - endpoints for 'forgot password' request
  and setting the new password


app/main/views/reset_password.py
```diff
@@ -0,0 +1,140 @@
...
+import six
+
+from flask import abort, current_app, flash, redirect, render_template, url_for, Markup
+
+from dmutils.user import User
+from dmutils.email import generate_token, decode_password_reset_token, send_email
+from dmutils.email.exceptions import EmailError
+
+from .. import main
+from ..forms.auth_forms import EmailAddressForm, ChangePasswordForm
```
TheDoubleK: Not for this PR, but there is now an `EmailAddressForm` in utils that we should move to using for all our email address form fields.

PR #6, opened 2017-09-14
Add a check for safe url for redirects after login

allait:
After the buyer users have been added to the list of roles the prefix
checks no longer restrict protocol-relative URLs (eg //example.com),
allowing the login page to redirect the user to an external domain.

This is adding a `is_safe_url` function based on the one from
http://flask.pocoo.org/snippets/62/ (it's also used by Notify).

The function is modified to return `False` for empty target URLs.

This also removes per-role redirect restrictions since their original
purpose was to prevent the open redirects. User role now only affects
the default redirect location if next_url was not set, otherwise
next_url is used as is.

### DIFF
--- a/tests/main/helpers/test_login_helpers.py
+++ b/tests/main/helpers/test_login_helpers.py
@@ -1,0 +1,62 @@
+import mock
+import pytest
+
+from collections import namedtuple
+
+from app.main.helpers.login_helpers import is_safe_url
+
+
+@pytest.fixture()
+def host_url(request):
+    fake_request = namedtuple('Request', ['host_url'])('http://localhost/')
+    request_patch = mock.patch('app.main.helpers.login_helpers.request', fake_request)
+    request.addfinalizer(request_patch.stop)
+
+    return request_patch.start()
+
+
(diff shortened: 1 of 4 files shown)


risicle:
LGTM :+1:

tests/main/helpers/test_login_helpers.py
```diff
@@ -0,0 +1,62 @@
...
+@pytest.mark.parametrize('next_url', [None, ''])
+def test_empty_next_url_is_not_safe(host_url, next_url):
+    assert not is_safe_url(next_url)
+
+
+@pytest.mark.parametrize('next_url', [
+    'https://example.com/suppliers',
+    'http://localhost:5000/',
+    'http://example.com/',
+    '//example.com/',
```
katstevens: Worth adding in the `@example.com` pattern as well (it was mentioned in that article)?

allait: That's a greate shout :+1:

I've tried it and it fails the test, but the reason is that the redirect URL ends up being `http://localhost/@example.com`, which is fine. I've added a test to `test_auth.py` for this case.

katstevens: Awesome - looks good!

PR #11, opened 2017-09-18
create_user template: modify displayed text

risicle:
Story https://trello.com/c/fNL5RPyB/81-update-the-create-account-page-to-remove-the-word-contributor-1

Now this looks like
new

### DIFF
--- a/app/templates/auth/create-user.html
+++ b/app/templates/auth/create-user.html
@@ -102,7 +102,7 @@
             {%
               with
               type = "save",
-              label = "Create account" if role == 'buyer' else "Save and continue"
+              label = "Create account"
             %}
               {% include "toolkit/button.html" %}
             {% endwith %}
--- a/app/templates/auth/create-user.html
+++ b/app/templates/auth/create-user.html
@@ -1,7 +1,7 @@
 {% extends "_base_page.html" %}
 
 {% block page_title %}
-  Create {% if role == 'supplier' %}contributor {% endif %}account – Digital Marketplace
+  Create account – Digital Marketplace
 {% endblock %}
 
 {% block main_content %}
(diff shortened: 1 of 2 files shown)


risicle:
Ok now it looks like this (action text changed as well as page "title" (not visible))
newer

app/templates/auth/create-user.html
```diff
@@ -105,7 +102,7 @@ <h3 class="validation-masthead-heading" id="validation-masthead-heading">
             {%
               with
               type = "save",
-              label = "Create account" if role == 'buyer' else "Create contributor account"
+              label = "Create account" if role == 'buyer' else "Save and continue"
```
allait: Why "Save and continue" for suppliers? This the final step for account creation in both cases, so "Create account" should make sense in both cases. Or are there some new additional supplier steps after this one?

risicle:
This was not my decision. See story on trello.

app/templates/auth/create-user.html
risicle: This was not my decision. See story on trello.

risicle: @constancecerf

constancecerf: If you are going from the create a supplier journey, users think they've already created the account and now they are just creating a login to activate it. So the other option would be to say either 'Complete account' or 'Confirm account'

constancecerf: I've just had a chat with Karl, and @risicle, we can just be consistent with what's already in the code as long as it makes sense here.

allait: @constancecerf so can we keep "Create account" for both buyers and suppliers?

PR #16, opened 2017-11-02
Pull in latest utils

MuriloDalRi:
The latest version of utils removes a bit of code that deals with the case where tokens don't have a user role. The new way of generating tokens should have made that code obsolete, so this PR *shouldn't* break anything.


MuriloDalRi:
The two tests failing are using the old ```generate_token``` function. I'll update them to use the new ```send_user_account_email```

PR #20, opened 2017-12-05
Change from flake8-putty to flake8-per-file-ignores

samuelhwilliams:
This allows us to use newer Python3 syntax including f-strings and type annotations.

```
samuelwilliams:~/git/digitalmarketplace-user-frontend [master]$ make test-flake8
[ -z $VIRTUAL_ENV ] && [ ! -d venv ] && virtualenv -p python3 venv || true
/Users/samuelwilliams/git/digitalmarketplace-user-frontend/venv/bin/flake8 .
./app/__init__.py:0:1: X100 Superfluous per-file-ignores for E402
make: *** [test-flake8] Error 1
```

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -9,7 +9,8 @@ exclude = venv*,__pycache__,node_modules,bower_components
 ignore = D203,W503
 max-complexity = 7
 max-line-length = 120
-putty-ignore =
-    **/__init__.py : +F401
-    app/__init__.py, app/main/__init__.py, app/status/__init__.py : +E402
-    app/main/views/*.py : +C901
+per-file-ignores =
+    **/__init__.py : F401
+    app/main/__init__.py : E402
+    app/status/__init__.py : E402
+    app/main/views/*.py : C901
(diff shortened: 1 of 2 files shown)


.flake8
```diff
@@ -9,7 +9,8 @@ exclude = venv*,__pycache__,node_modules,bower_components
...
 max-complexity = 7
 max-line-length = 120
-putty-ignore =
-    **/__init__.py : +F401
-    app/__init__.py, app/main/__init__.py, app/status/__init__.py : +E402
-    app/main/views/*.py : +C901
+per-file-ignores =
+    **/__init__.py : F401
+    app/main/__init__.py : E402
+    app/status/__init__.py : E402
```
katstevens: Is this missing the `app/__init__.py: E402`, or has that been fixed elsewhere?

samuelhwilliams: `per-file-ignores` reports a custom flake8 error (X100) if you specify an error to ignore for a file and said file doesn't trigger that error. `putty-flake8` seems to just happily accept those. So when we changed extensions, we got notified of extraneous errors, which I've removed. I've updated the PR again with the original report after changing extensions. :)

PR #33, opened 2018-01-17
Adds in /help page

DilwoarH:
Requires:
https://github.com/alphagov/digitalmarketplace-utils/pull/346


DilwoarH:
@idavidmcdonald - it seems that user frontend has its own external file :\

app/external/views/external.py
```diff
@@ -13,6 +13,11 @@ def suppliers_list_by_prefix():
     raise NotImplementedError()
 
 
+@external.route('/help')
```
samuelhwilliams: Is there anything stopping us migrating these externals to utils and then pulling them in so that we centralise this anomaly?

DilwoarH: I wasn't sure why it had its own external file. I assumed it was intended for some purpose.

samuelhwilliams: I think it was probably just not migrated/thought of at the time. It would be good to do it now if you're happy to?

PR #35, opened 2018-02-07
Use `SHARED_EMAIL_KEY` to encrypt reset password email tokens

idavidmcdonald:
https://trello.com/c/WTbemGdn/219-reset-password-emails-are-using-secretkey-instead-of-sharedemailkey-to-encrypt-tokens

Note, tests will likely fail until utils PR - https://github.com/alphagov/digitalmarketplace-utils/pull/354 - is merged and then pulled into this pull request.

We will be changing our tokens for reset password emails to be
generated using the `SHARED_EMAIL_KEY` rather than the `SECRET_KEY`.
This is because
- this will be consistent with the other email tokens we generate
- although there is technically no harm in using `SECRET_KEY`,
the actual purpose of the `SECRET_KEY` is to be used for encrypting
Flask sessions. This would mean if we needed to rotate a key for
our Flask sessions it would also break out email tokens and vice
versa. By using different keys we should reduce the risk of a
token rotation having an impact on too large a part of the site

Note, the new version of utils brought in will mean that we still
succesfully decrypt and allow tokens that were previously
generated using the old `SECRET_KEY`. After 1 full day of support,
they would all become invalid anyway so after that point we can
also remove support for allowing old `SECRET_KEY` tokens.

### DIFF
--- a/config.py
+++ b/config.py
@@ -74,7 +74,7 @@ class Test(Config):
 
     DM_NOTIFY_API_KEY = "not_a_real_key-00000000-fake-uuid-0000-000000000000"
     SHARED_EMAIL_KEY = "KEY"
-    SECRET_KEY = "KEY"
+    SECRET_KEY = "KEY2"
 
 
 class Development(Config):
--- a/app/main/views/reset_password.py
+++ b/app/main/views/reset_password.py
@@ -49,7 +49,7 @@ def send_reset_password_email():
                 {
                     "user": user.id
                 },
-                current_app.config['SECRET_KEY'],
+                current_app.config['SHARED_EMAIL_KEY'],
                 current_app.config['RESET_PASSWORD_SALT']
             )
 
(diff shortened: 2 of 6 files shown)


config.py
```diff
@@ -74,7 +74,7 @@ class Test(Config):
 
     DM_NOTIFY_API_KEY = "not_a_real_key-00000000-fake-uuid-0000-000000000000"
     SHARED_EMAIL_KEY = "KEY"
-    SECRET_KEY = "KEY"
+    SECRET_KEY = "KEY2"
```
Wynndow: I'm guessing this is just to differentiate between the two keys for tests? Make sure no confusion?

idavidmcdonald: Yes in that there should be no confusion in a developers understanding that the two are different but also so that our tests don't risk being incorrect. We don't want our
tests accidentally passing if we were accidentally using one of the keys instead of the other but we wouldn't know because their values are the same.

PR #37, opened 2018-02-13
Revert back to eea/odfpy

samuelhwilliams:
Update utils to use eea/odfpy rather than our interim alphagov/odfpy
fork, which we created to allow us to patch a bug. Now that the bug has
been accepted into the main library as of release 1.3.6
(https://github.com/eea/odfpy/pull/72) we no longer need to support
own our fork.

Also adds a comment when registering the external blueprint from 
dm-utils to ensure it is registered last.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -32,6 +32,9 @@ def create_app(config_name):
     from .main import main as main_blueprint
 
     application.register_blueprint(main_blueprint, url_prefix='/user')
+
+    # Must be registered last so that any routes declared in the app are registered first (i.e. take precedence over
+    # the external NotImplemented routes in the dm-utils external blueprint).
     application.register_blueprint(external_blueprint)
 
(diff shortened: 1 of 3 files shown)


app/__init__.py
```diff
@@ -32,6 +32,9 @@ def create_app(config_name):
     from .main import main as main_blueprint
 
     application.register_blueprint(main_blueprint, url_prefix='/user')
+
+    # Must be registered last so that any routes declared in the app are registered first (i.e. take precedence over
+    # the external NotImplemented routes in the dm-utils external blueprint).
```
risicle: #random

samuelhwilliams: I added it to all apps in the same place after registering external blueprints early caused things to break. See https://github.com/alphagov/digitalmarketplace-briefs-frontend/pull/77

PR #43, opened 2018-03-09
Backend form for sign up to mailing list

DilwoarH:
As a developer I need to be able to use a form for validating responses to the sign me up for user research page.

This form should follow existing pattens so it is easily extensible.

See prototype for fields and help text

Ticket: https://trello.com/c/qoRyFSEn/130-backend-form-for-sign-up-to-mailing-list

screen shot 2018-03-13 at 14 47 44

### DIFF
--- a/app/main/views/notifications.py
+++ b/app/main/views/notifications.py
@@ -1,0 +1,44 @@
+# coding: utf-8
+from flask_login import (current_user, login_required)
+from flask import (
+    flash,
+    render_template,
+    request,
+    url_for
+)
+from .. import main
+from ... import data_api_client
+
+
+@main.route('/notifications/user-research', methods=["GET", "POST"])
+@login_required
+def user_research_consent():
+    if request.method == "POST":
+        userResearchOptIn = request.form.get('user_research_opt_in') == "True"
+        data_api_client.update_user(
+            current_user.id,
+            user_research_opted_in=userResearchOptIn,
+            updater=current_user.email_address
+        )
+
+        flash("Your preference has been saved", 'success')
+    else:
+        user = data_api_client.get_user(current_user.id)
+        userResearchOptIn = user['users']['userResearchOptedIn']
+
+    if current_user.role == 'supplier':
+        dashboard_url = url_for('external.supplier_dashboard')
+    else:
+        dashboard_url = url_for('external.buyer_dashboard')
+
+    additional_headers = []
+    cookie_name = 'seen_user_research_message'
+
+    if cookie_name not in request.cookies:
+        additional_headers = {'Set-Cookie': "{}=yes; Path=/".format(cookie_name)}
+
+    return render_template(
+        "notifications/user-research-consent.html",
+        userResearchOptIn=userResearchOptIn,
+        dashboard_url=dashboard_url
+    ), 200, additional_headers
--- a/app/templates/notifications/user-research-consent.html
+++ b/app/templates/notifications/user-research-consent.html
@@ -1,0 +1,89 @@
+{% extends "_base_page.html" %}
+
+{% block page_title %}User reserch notification – Digital Marketplace{% endblock %}
+
+{% block breadcrumb %}
+  {%
+    with items = [
+      {
+        "link": "/",
+        "label": "Digital Marketplace"
+      },
+      {
+        "link": dashboard_url,
+        "label": "Your account"
+      }
+    ]
+  %}
+    {% include "toolkit/breadcrumb.html" %}
+  {% endwith %}
+{% endblock %}
+
+{% block main_content %}
+
+<div class="grid-row notification-user-research">
+
+  <div class="column-two-thirds">
+    {% with heading = "{}".format(
+        "Join the user research mailing list" 
+          if not userResearchOptIn
+          else "Unsubscribe from the user research mailing list"
+        ) %}
+      {% include "toolkit/page-heading.html" %}
+    {% endwith %}
+
+    {% if not userResearchOptIn %}
+      <p class="text">
+        Help us improve the Digital Marketplace by taking part in user research.
+      </p>
+      <p class="text">
+        You don’t have to take part just because you’ve signed up and you can unsubscribe at any time.
+      </p>
+    {% else %}
+      <p class="text">
+         You can join again at any time.
+      </p>
+    {% endif %}
+
+    <form action="{{ url_for('main.user_research_consent') }}" method="POST">
+      <input type="hidden" name="csrf_token" value="{{ csrf_token() }}" />
+      {%
+        with
+        type = "checkbox",
+        name = "user_research_opt_in",
+        value = [userResearchOptIn],
+        options = [
+          { 
+            "label": "Send me emails about opportunities to get involved in user research",
+            "value": True,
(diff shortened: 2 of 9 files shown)


MuriloDalRi:
It would be nice to have a bit more information about why we want people to sign up for user research, what is it, why we do it, what are we going to ask from them, etc. @constancecerf thoughts?

Also, and this is probably out of scope for this mission but we should make a note, we might want to do user research with admin users in the future and it would be good if they could sign up as well.

app/templates/notifications/user-research-consent.html
```diff
@@ -0,0 +1,72 @@
...
+          { 
+            "label": "Send me emails about opportunities to get involved in user research",
+            "value": True,
+          }
+        ]
+      %}
+        {% include "toolkit/forms/selection-buttons.html" %}
+      {% endwith %}
+
+      {% with label="Save changes", type="save" %}
```
benvand: This button should get a `name` argument so we can find it with the functional tests

app/templates/notifications/user-research-consent.html
```diff
@@ -0,0 +1,72 @@
...
+      You don’t have to take part just because you’ve signed up and you can unsubscribe at any time.
+    </p>
+
+    <form action="{{ url_for('main.user_research_consent') }}" method="POST">
+      <input type="hidden" name="csrf_token" value="{{ csrf_token() }}" />
+      {%
+        with
+        type = "checkbox",
+        name = "user_research_opt_in",
+        value = [userResearchOptIn],
```
benvand: You should be able to get this value from `current_user` which is made available in all templates. That way you can have one less template argument.

app/main/views/notifications.py
```diff
@@ -0,0 +1,39 @@
...
+        user = data_api_client.get_user(current_user.id)
+        userResearchOptIn = user['users']['userResearchOptedIn']
+
+    if current_user.role == 'supplier':
+        dashboard_url = url_for('external.supplier_dashboard')
+    else:
+        dashboard_url = url_for('external.buyer_dashboard')
+
+    return render_template("notifications/user-research-consent.html",
+                           userResearchOptIn=userResearchOptIn,
```
benvand: This function call is formatted a bit weirdly.
Here you've aligned the kwargs with the opening delimiter, but further up in the call to `update_user` they're indented at 4 spaces.
```
my_function(foo,
                     bar,
                      baz)
```
```
my_function(
    foo,
    bar,
    baz
)
```
It was probably your editor, mine does that too sometimes. I personally prefer the second way. It keeps spacing standard at multiples of 4. Whereas in this function call the kwargs are indented to 27 spaces.

Regardless, it's best to pick one and stick with it.

benvand:
Good stuff overall!

I think there's an argument that using a WTForm would have been a bit more extensible, allow for passing of errors and passing current values.
See the InviteAdminForm (form), invite_admin_user (view) and invite_admin_user.html (template) for a good example.

It's up to you if you want to implement it that way. In this situation it doesn't make a great deal of difference because there are no errors. But it wouldn't be a great deal of work and it might be good to work with the form classes if you haven't before.

:+1:

app/templates/notifications/user-research-consent.html
DilwoarH: does this value update on refresh each time? ie. If I opt in, on the next view, will it be true?

katstevens: The Flask user object is populated via an API call on login. If you opt in during the session I don't think it will update the user object until you log out/log back in again.

DilwoarH: Have checked, there is a small delay before this value is updated.

DilwoarH:
@benvand - I was thinking of using it but because of the reasons you stated (ie. no errors), I thought it wasn't worth adding an extra level of complexity. I have updated the code with the changes requested except for one.

app/main/views/notifications.py
```diff
@@ -0,0 +1,41 @@
...
+from ... import data_api_client
+
+
+@main.route('/notifications/user-research', methods=["GET", "POST"])
+def user_research_consent():
+    if current_user.is_anonymous():
+        return redirect(url_for('main.render_login', next=url_for('main.user_research_consent')))
+
+    if request.method == "POST":
+        userResearchOptIn = True if request.form.get('user_research_opt_in') == "True" else False
```
benvand: This statement can be shortened to:
```
userResearchOptIn = request.form.get('user_research_opt_in') == "True"
```

app/main/views/notifications.py
```diff
@@ -0,0 +1,41 @@
...
+    request,
+    url_for
+)
+from .. import main
+from ... import data_api_client
+
+
+@main.route('/notifications/user-research', methods=["GET", "POST"])
+def user_research_consent():
+    if current_user.is_anonymous():
```
benvand: Can we not use `from flask_login import login_required` to achieve this behaviour out of the box?

benvand:
We'll need some tests for this to check the logic in the template and the access rights.

DilwoarH:
hey ben tests added

app/main/views/notifications.py
```diff
@@ -0,0 +1,47 @@
...
+        dashboard_url = url_for('external.buyer_dashboard')
+
+    additional_headers = []
+    cookie_name = 'seen_user_research_message'
+
+    if cookie_name not in request.cookies:
+        cookie_expiry = datetime.datetime.now() + datetime.timedelta(30)
+        cookie_expiry = cookie_expiry.strftime("%a, %d %b %Y %H:%M:%S GMT")
+        additional_headers = {'Set-Cookie': "{}=yes; Expires={}; Path=/".format(cookie_name, cookie_expiry)}
+
```
DilwoarH: @benvand cookies now being added when they don't exist

TheDoubleK:
I know I'm late to the party here but I think this should definitely use a WTForm and it should be added before we release this if possible.

There are two reasons I say this:
1. All forms should be validated and POST routes should return an error if they are sent unexpected data.  As it stands this form will happily take data that should *never* be sent from our frontend form and set data on the user (for opt-in to be False). e.g. a POST with `user_research_opt_in=WhatIsThisValueDoingHere` or without any value for `user_research_opt_in` at all.

A simple WTForm with one checkbox would catch these and return a validation error on the form.

For example, a test something like this should pass:
```
@pytest.mark.parametrize('bad_value', [
    '',
    'SomethingNotABoolean',
    ['a','list'],
    ['True'],
])
def test_post_errors_for_invalid_data(self, bad_value):
    res = self.client.post("/users/notifications/user-research", data={"user_research_opt_in": bad_value})
    assert res.status_code=400
```
2. Consistency.  We use WTForms for in-app form pages like this everywhere else we have them (or, at least, should do).  By treating all forms in a consistent way it means future developers have patterns to follow when building out new features. It may feel safe to omit the WTForm in this case, but then when someone comes along later and needs to build a form with a wider range of failure cases if they fnd this one and follow the pattern set here they'll think we "roll our own" form processing, which is something we really don't want to encourage.

DilwoarH:
@TheDoubleK - thanks, I'll get that updated.

PR #45, opened 2018-03-13
Analytics JS cleanup for User FE

katstevens:
Background: the custom dimensions work (https://trello.com/c/Jj4cz4u4/336-implement-a-custom-dimension-for-buyer-supplier-and-admin-in-ga) consolidated the analytics code into the FE toolkit, but that story did not include updating the User FE (as no post-login custom dimensions could be set there). 

Based on the equivalent PR for Briefs FE: https://github.com/alphagov/digitalmarketplace-briefs-frontend/pull/81

- Bumped to the latest toolkit version 
- Pulls in the shared analytics JS for use in the User FE
- Updates the virtual page view tests to bring them in line with the FE toolkit VPV code


spec/javascripts/unit/AnalyticsSpec.js
```diff
@@ -70,30 +70,29 @@ describe("GOVUK.Analytics", function () {
     it("Should not call google analytics without a url", function () {
       $analyticsString = $("<div data-analytics='trackPageView'/>");
       $(document.body).append($analyticsString);
-      window.GOVUK.GDM.analytics.virtualPageViews.init();
+      window.GOVUK.GDM.analytics.virtualPageViews();
```
galund: I'm not sure I understand why these test cases are even in this application? Presumably they duplicate some in another app, that actually uses VPVs?!

katstevens: Yep, I think they were just copied over. For now they're useful to check that the toolkit is being pulled in correctly. We don't have any analytics tests in the toolkit yet, that was going to be the next tech debt step.

galund: I'm sure they're _that_ useful, as long as the toolkit's being tested _somewhere_ that'd be fine by me. I'd prefer to remove the duplicate code unless there's a strong concern that it needs to be here.

katstevens: More than happy to take them out (I've satisfied myself that the toolkit is working ok for this app).

katstevens:
I've taken out all the duplicated JS tests, and added a note in the README saying there's not much point in running `test-javascript` right now.

PR #49, opened 2018-03-21
WTForm for user research opt in

benvand:
Use a WTform for the user research opt int as per @TheDoubleK s comments
Adding a couple of tests now

### DIFF
--- a/tests/main/views/test_notifications.py
+++ b/tests/main/views/test_notifications.py
@@ -53,3 +53,21 @@ class TestUserResearchNotifications(BaseApplicationTest):
         res = self.client.get("/user/notifications/user-research")
         assert res.status_code == 200
         assert "Unsubscribe from the user research mailing list" in res.get_data(as_text=True)
+
+    @mock.patch('app.main.views.notifications.data_api_client', autospec=True)
+    def test_user_research_opt_in(self, data_api_client):
+        self.login_as_buyer()
+        self.client.post("/user/notifications/user-research", data={"user_research_opt_in": "True"})
+        self.assert_flashes('Your preference has been saved', expected_category='success')
+        assert data_api_client.update_user.call_args_list == [
+            mock.call(123, updater='buyer@email.com', user_research_opted_in=True)
+        ]
+
+    @mock.patch('app.main.views.notifications.data_api_client', autospec=True)
+    def test_user_research_opt_out(self, data_api_client):
+        self.login_as_buyer()
+        self.client.post("/user/notifications/user-research", data={})
+        self.assert_flashes('Your preference has been saved', expected_category='success')
+        assert data_api_client.update_user.call_args_list == [
+            mock.call(123, updater='buyer@email.com', user_research_opted_in=False)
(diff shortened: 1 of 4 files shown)


TheDoubleK:
Just trying this out locally - this version of the page seems really confusing to me... the heading says "Unsubscribe" but the box is already ticked - I really have to think and read all the words to understand that I need to untick the box before hitting the button to actually unsubscribe.

screen shot 2018-03-22 at 11 16 09

TheDoubleK:
Also, is there a banner message I should see that links to this page?  I haven't seen it anywhere when logging in as a supplier locally - I had to navigate to the page directly to try it out.

tests/main/views/test_notifications.py
```diff
@@ -53,3 +53,21 @@ def test_unsubscribe_content(self):
...
+        self.assert_flashes('Your preference has been saved', expected_category='success')
+        assert data_api_client.update_user.call_args_list == [
+            mock.call(123, updater='buyer@email.com', user_research_opted_in=True)
+        ]
+
+    @mock.patch('app.main.views.notifications.data_api_client', autospec=True)
+    def test_user_research_opt_out(self, data_api_client):
+        self.login_as_buyer()
+        self.client.post("/user/notifications/user-research", data={})
+        self.assert_flashes('Your preference has been saved', expected_category='success')
```
DilwoarH: @TheDoubleK - it now redirects to the dashboards, as it's going to a different app I've had to assert the flash messages using this function. Is this the correct way?

katstevens: This looks fine, however we're in the process of revamping the flash messages across all apps so you might want to look at the latest master of the Supplier FE to check how it will be displayed.

katstevens: (Ah - it's not on master yet! See https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/845)

benvand:
@TheDoubleK @katstevens The banners for buyers and suppliers are in 
https://github.com/alphagov/digitalmarketplace-briefs-frontend/pull/88
https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/832/files
respectively. 

We'll have a look and double check in QA but the flash messages are just being flashed and displayed using the existing mechanisms. Ie the user-frontend view uses `flash("Your preference has been saved", 'success')` and that is picked up in the supplier frontend/ briefs frontend with the standard `{% with messages = get_flashed_messages(with_categories=true, category_filter=["error", "success"]) %}` block.

benvand:
@TheDoubleK 
> have just one version of the page that works for both cases, and then the meaning of the box being ticked would be more obvious

I think the above is a good idea, and we could have used something like the below:

screen shot 2018-03-27 at 10 58 38

I'm going to have to hold my hands up and say I completely forgot that we had this pattern. I'll bring it up with design but I'd like to forge on with this as an MVP. Maybe design can redo the mock ups in the firebreak if they want to make it clearer 🤷‍♂️

PR #53, opened 2018-03-28
notify example redirection: switch the target email addresses to SES's equivalent simulated addresses

risicle:
This should work fine as long as notify is using SES as their backend. if they change this, we just have to change this setting. using this has the advantage that it will appear to be "delivered" as far as notify's concerned and a record of the message's reference id will be kept.

### DIFF
--- a/config.py
+++ b/config.py
@@ -98,8 +98,8 @@ class Live(Config):
 
     # use of invalid email addresses with live api keys annoys Notify
     DM_NOTIFY_REDIRECT_DOMAINS_TO_ADDRESS = {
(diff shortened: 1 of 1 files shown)


katstevens:
Once all this lands, can we make a note of it in the manual so it's 💯clear which environments are expected send emails & whether we expect them to arrive?

idavidmcdonald:
There is a section for this in the manual already but worth double checking it/adding to it: https://alphagov.github.io/digitalmarketplace-manual/emails.html

risicle:
Yes

- [ ] TODO update documentation

PR #54, opened 2018-03-29
414 inadequate session termination

katstevens:
Trello https://trello.com/c/NStzgIbd/414-inadequate-session-termination

The recent pen test highlighted that our sessions are not invalidated properly on logout. 

This possible solution sets the session cookie's expiry to the Unix epoch (which one would hope is in the past). 

I'm not 100% sure this will be sufficient! It's also hard to test in-browser as a new session cookie is immediately set for the logged out user. 

Thoughts/alternative suggestions welcome.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -7,10 +7,9 @@
 # W503: line break before binary operator
 exclude = venv*,__pycache__,node_modules
 ignore = D203,W503
-max-complexity = 7
+max-complexity = 9
 max-line-length = 120
 per-file-ignores =
     **/__init__.py : F401
     app/main/__init__.py : E402
     app/status/__init__.py : E402
-    app/main/views/*.py : C901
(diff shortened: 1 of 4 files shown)


katstevens:
Yeah... as discussed just now, I don't think we store the sessions server side (at least not anywhere I know about). [Flask-Session](https://pythonhosted.org/Flask-Session/) allows you to configure a DB table or Redis cache to store things. But it feels a bit nuts introducing a way of storing sessions just so we can delete them.

risicle:
Yeah it's not *actually* possible to do an ad-hoc invalidation of this type of signed session cookie - one of the reasons I don't like them. You only have that ability if you use server-side sessions. A malicious user would just ignore the cookie in this response.

Introducing server-side sessions would probably involve adding another moving part for us because of the wonderful way our existing stuff is designed with our database hidden behind an "api"...

PR #56, opened 2018-04-05
Refactor User FE tests

katstevens:
Following on from the Briefs FE test refactor (https://github.com/alphagov/digitalmarketplace-briefs-frontend/pull/93), refactoring to standardise test classes as follows:

- remove unnecessary `self.app.app_context` contexts (not required for FE app tests)
- `super(TestClassName, self)` ---> `super()` for Python 3
- patch Data API client at setup/teardown level instead of decorators, to tidy parameters and avoid accidentally patching a single API method rather than the whole API.
- ensuring `teardown()` methods include `super()` 

The `TestUserResearchNotifications` class uses functions that import the API from two modules (`auth.py` and `notifications.py`), so two patches are required.

### DIFF
--- a/tests/main/views/test_auth.py
+++ b/tests/main/views/test_auth.py
@@ -14,19 +14,16 @@ PASSWORD_EMPTY_ERROR = "You must provide your password"
 class TestLogin(BaseApplicationTest):
 
     def setup_method(self, method):
-        super(TestLogin, self).setup_method(method)
-
-        data_api_client_config = {'authenticate_user.return_value': self.user(
-            123, "email@email.com", 1234, 'name', 'name'
-        )}
-
-        self._data_api_client = mock.patch(
-            'app.main.views.auth.data_api_client', **data_api_client_config
+        super().setup_method(method)
+        self.data_api_client_patch = mock.patch('app.main.views.auth.data_api_client', autospec=True)
+        self.data_api_client = self.data_api_client_patch.start()
+        self.data_api_client.authenticate_user.return_value = self.user(
+            123, "email@email.com", 1234, 'name', 'name'  # supplier user
         )
-        self.data_api_client_mock = self._data_api_client.start()
 
     def teardown_method(self, method):
-        self._data_api_client.stop()
+        self.data_api_client_patch.stop()
+        super().teardown_method(method)
 
     def test_should_show_login_page(self):
         res = self.client.get("/user/login")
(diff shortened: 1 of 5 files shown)


tests/main/views/test_auth.py
```diff
@@ -14,19 +14,16 @@
...
 
     def setup_method(self, method):
-        super(TestLogin, self).setup_method(method)
-
-        data_api_client_config = {'authenticate_user.return_value': self.user(
-            123, "email@email.com", 1234, 'name', 'name'
-        )}
-
-        self._data_api_client = mock.patch(
-            'app.main.views.auth.data_api_client', **data_api_client_config
```
CrystalPea: So we do not need `**data_api_client_config` anymore? Just curious :).

tests/main/views/test_auth.py
```diff
@@ -42,17 +39,15 @@ def test_should_redirect_to_supplier_dashboard_on_supplier_login(self):
         assert res.location == 'http://localhost/suppliers'
         assert 'Secure;' in res.headers['Set-Cookie']
 
-    @mock.patch('app.main.views.auth.data_api_client')
-    def test_should_redirect_to_homepage_on_buyer_login(self, data_api_client):
-        with self.app.app_context():
```
CrystalPea: that's interesting, why did we need this app context before, but we do not need it anymore? Anything to do with our refactor or just how new versions of our stack work?

katstevens: Ben gave a good explanation of this on the equivalent Briefs FE refactoring PR (https://github.com/alphagov/digitalmarketplace-briefs-frontend/pull/93):

```We shouldn't ever need a self.app.app_context call in the frontend tests. IIRC a call to a frontend view should create a new app and request context for the given request.```

This test is doing a simple `self.client.post(...)` - the app context should be created as part of the client's post request.

We should only occasionally need the `self.app.app_context` inside the test, usually when we want to test how code behaves under a different app config setting. For example in the Buyer FE there's a test where we override `current_app.config['DM_SEARCH_PAGE_SIZE']` - we only want to change it for the scope of that particular test.

katstevens: (I don't know why it was there before - probably a copy and paste from somewhere where we did need it?)

tests/main/views/test_auth.py
katstevens: Nope - I explicitly set the `authenticate_user.return_value` below rather than passing it as args to the patch. There's no real difference between the two, just being consistent with styling in the other apps.

CrystalPea: cool, thanks for the explanation :). I guess also better to have it explicit, more readable :3.

tests/main/views/test_auth.py
CrystalPea: Ahhh, ok, thank you 💖

PR #58, opened 2018-04-27
Enable users to change their password when logged in

CrystalPea:
Following our security report recommendation (see ticket for link to report)
The Trello ticket: https://trello.com/c/TieWrBG5/44-password-reset-link-from-logged-in-session

screen shot 2018-04-27 at 10 57 06

### DIFF
--- a/app/main/forms/auth_forms.py
+++ b/app/main/forms/auth_forms.py
@@ -32,7 +32,7 @@ class EmailAddressForm(FlaskForm):
     )
 
 
-class ChangePasswordForm(FlaskForm):
+class PasswordResetForm(FlaskForm):
     password = PasswordField(
         'New password', id="input_password",
         validators=[
@@ -52,7 +52,10 @@ class ChangePasswordForm(FlaskForm):
     )
 
 
-class ChangeOldPasswordForm(ChangePasswordForm):
+class PasswordChangeForm(PasswordResetForm):
+    """
+    Subclasses PasswordResetForm so we can keep validation/password policy in one place
+    """
     old_password = PasswordField(
         'Old password', id="input_old_password",
         validators=[
--- a/app/main/forms/auth_forms.py
+++ b/app/main/forms/auth_forms.py
@@ -34,7 +34,7 @@ class EmailAddressForm(FlaskForm):
 
 class ChangePasswordForm(FlaskForm):
     password = PasswordField(
-        'Password', id="input_password",
+        'New password', id="input_password",
         validators=[
             DataRequired(message="You must enter a new password"),
             Length(min=10,
@@ -44,10 +44,19 @@ class ChangePasswordForm(FlaskForm):
         ]
     )
     confirm_password = PasswordField(
-        'Confirm password', id="input_confirm_password",
+        'Confirm new password', id="input_confirm_password",
         validators=[
             DataRequired(message="Please confirm your new password"),
             EqualTo('password', message="The passwords you entered do not match")
+        ]
+    )
+
+
+class ChangeOldPasswordForm(ChangePasswordForm):
+    old_password = PasswordField(
+        'Old password', id="input_old_password",
+        validators=[
+            DataRequired(message="You must enter your old password"),
         ]
     )
 
--- a/app/main/views/reset_password.py
+++ b/app/main/views/reset_password.py
@@ -9,6 +9,7 @@ from dmutils.email.helpers import hash_string
 
 from .. import main
 from ..forms.auth_forms import EmailAddressForm, ChangePasswordForm, ChangeOldPasswordForm
+from ..helpers.login_helpers import get_user_dashboard_url
 from ... import data_api_client
 
 
@@ -146,14 +147,7 @@ def update_password(token):
 @login_required
 def change_password():
     form = ChangeOldPasswordForm()
-    if current_user.role == 'supplier':
-        dashboard_url = url_for('external.supplier_dashboard')
-    elif current_user.role == "buyer":
-        dashboard_url = url_for('external.buyer_dashboard')
-    elif current_user.role.startswith('admin'):
-        dashboard_url = url_for('external.admin_dashboard')
-    else:
-        dashboard_url = url_for('external.index')
+    dashboard_url = get_user_dashboard_url(current_user)
 
     if request.method == 'POST':
         if form.validate_on_submit():
--- a/app/main/views/reset_password.py
+++ b/app/main/views/reset_password.py
@@ -8,7 +8,7 @@ from dmutils.email import DMNotifyClient, generate_token, decode_password_reset_
 from dmutils.email.helpers import hash_string
 
 from .. import main
-from ..forms.auth_forms import EmailAddressForm, ChangePasswordForm, ChangeOldPasswordForm
+from ..forms.auth_forms import EmailAddressForm, PasswordResetForm, PasswordChangeForm
 from ..helpers.login_helpers import get_user_dashboard_url
 from ... import data_api_client
 
@@ -111,13 +111,13 @@ def reset_password(token):
 
     return render_template("auth/reset-password.html",
                            email_address=email_address,
-                           form=ChangePasswordForm(),
+                           form=PasswordResetForm(),
                            token=token), 200
 
 
 @main.route('/reset-password/<token>', methods=["POST"])
 def update_password(token):
-    form = ChangePasswordForm()
+    form = PasswordResetForm()
     decoded = decode_password_reset_token(token, data_api_client)
     if 'error' in decoded:
         flash(EXPIRED_PASSWORD_RESET_TOKEN_MESSAGE, 'error')
@@ -146,7 +146,7 @@ def update_password(token):
 @main.route('/change-password', methods=["GET", "POST"])
 @login_required
 def change_password():
-    form = ChangeOldPasswordForm()
+    form = PasswordChangeForm()
     dashboard_url = get_user_dashboard_url(current_user)
 
     if request.method == 'POST':
@@ -222,6 +222,6 @@ def change_password():
     else:
         return render_template(
             "auth/change-password.html",
-            form=ChangeOldPasswordForm(),
+            form=PasswordChangeForm(),
             dashboard_url=dashboard_url
         ), 200
--- a/app/main/views/reset_password.py
+++ b/app/main/views/reset_password.py
@@ -163,6 +163,9 @@ def change_password():
                         "User {user_id} successfully changed their password",
                         extra={'user_id': current_user.id}
                     )
+                    flash(PASSWORD_UPDATED_MESSAGE)
+                else:
+                    flash(PASSWORD_NOT_UPDATED_MESSAGE, 'error')
                 return redirect(url_for('external.supplier_dashboard'))
             else:
                 current_app.logger.info(
--- a/app/main/views/reset_password.py
+++ b/app/main/views/reset_password.py
@@ -166,6 +166,46 @@ def change_password():
                         "User {user_id} successfully changed their password",
                         extra={'user_id': current_user.id}
                     )
+
+                    notify_client = DMNotifyClient(current_app.config['DM_NOTIFY_API_KEY'])
+
+                    token = generate_token(
+                        {
+                            "user": current_user.id
+                        },
+                        current_app.config['SHARED_EMAIL_KEY'],
+                        current_app.config['RESET_PASSWORD_SALT']
+                    )
+
+                    try:
+                        notify_client.send_email(
+                            current_user.email_address,
+                            template_id=current_app.config['NOTIFY_TEMPLATES']['change_password_alert'],
+                            personalisation={
+                                'url': url_for('main.reset_password', token=token, _external=True),
+                            },
+                            reference='change-password-alert-{}'.format(hash_string(current_user.email_address))
+                        )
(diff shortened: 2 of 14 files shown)


app/main/views/reset_password.py
```diff
@@ -139,3 +140,54 @@ def update_password(token):
...
+def change_password():
+    form = ChangeOldPasswordForm()
+    if current_user.role == 'supplier':
+        dashboard_url = url_for('external.supplier_dashboard')
+    elif current_user.role == "buyer":
+        dashboard_url = url_for('external.buyer_dashboard')
+    elif current_user.role.startswith('admin'):
+        dashboard_url = url_for('external.admin_dashboard')
+    else:
+        dashboard_url = url_for('external.index')
```
Wynndow: What's the `else` here? Anonymous users? Or is this a path that in theory should never be touched?

app/main/views/reset_password.py
```diff
@@ -139,3 +140,54 @@ def update_password(token):
...
+    if request.method == 'POST':
+        if form.validate_on_submit():
+            # Make sure old password is valid
+            user_json = data_api_client.authenticate_user(
+                current_user.email_address,
+                form.old_password.data)
+            if user_json:
+                response = data_api_client.update_user_password(
+                    current_user.id,
+                    form.password.data,
```
Wynndow: I know this is how we do it when logging in, but it feels a bit weird to send the password in plain text like this. Doesn't it risk dropping out in a log file someone where if the request fails for some reason?

app/main/views/reset_password.py
```diff
@@ -139,3 +140,54 @@ def update_password(token):
...
+                    updater=current_user.email_address
+                )
+                if response:
```
Wynndow: Is just checking `response` here enough? Should you check the status code perhaps to make sure it's a 200? If the responses was a 400 or a 500 would this evaluate to false?

app/main/views/reset_password.py
Wynndow: This form only has `old_password ` doesn't it?

app/templates/auth/change-password.html
```diff
@@ -0,0 +1,122 @@
...
+                    </p>
+                    {% endif %}
+                    {{ form.old_password(class="text-box", autocomplete="off") }}
+                </div>
+            {% if form.old_password.errors %}
+                </div>
+            {% endif %}
+
+
+            {% if form.password.errors %}
```
Wynndow: This template only gets passed a `ChangeOldPasswordForm` which only has form.old_password I think.

Wynndow: Is this template replacing the `reset-passord.html` template? If so, does `reset_password.py` need updating?

app/main/views/reset_password.py
katstevens: Ah - I was going to refactor this! There's a similar bit of code in the user research view which I don't think checks for `admin` (which we do here, obviously).

app/main/views/reset_password.py
katstevens: No - it inherits from `ChangePasswordForm` which has `password` and `confirm_password`.

katstevens: The api client catches any `HTTPError`s and returns `False` (hence just checking for 'response' below) without logging. I think we're probably OK?

app/main/views/reset_password.py
katstevens: See above - the API client handles this for us.

app/main/views/reset_password.py
Wynndow: ah nice!

app/templates/auth/change-password.html
```diff
@@ -0,0 +1,122 @@
...
+      }
+    ]
+  %}
+    {% include "toolkit/breadcrumb.html" %}
+  {% endwith %}
+{% endblock %}
+
+{% block main_content %}
+
+{% if form.errors|length > 1 %}
```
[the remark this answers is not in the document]
katstevens: OK - this pattern is used across the app as a 'there's multiple things wrong with this form' message, with bookmark links to jump to that error in the form. 

A single error would still be shown at field level, but it feels a bit like overkill to have two messages for one error? I can check with a designer in case this pattern usage has changed?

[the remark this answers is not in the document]
Wynndow: I think this is a thing we bumped into when doing the supplier details stuff in the last mission. The word from Laurence was that for consistency that even on a single question page we had the validation masthead as well as the field level thingy.

katstevens:
@Wynndow I think all the comments have been addressed now.

galund:
I'm fine with this I think, but just so the argument has definitely been rehearsed: one of the reasons we proposed just having "reset password" _instead_ of "change password" was (apart from it would be less work to implement, which is now moot) that it had the benefit of ensuring the user *did* still have (sole) access to their email account.

So you couldn't get access to someone's account and then change the password, without an email notification going out.

The danger here is that someone could have their password changed without any email notification going out (unless I've missed where you do that), and so be unaware that their account had been breached.

katstevens:
It's a good question - it wasn't specifically mentioned in the pen test recommendations, but we were considering sending an email notification along the lines of "your password has been changed", which shouldn't be too hard to do (with some help from a friendly content person). Would releasing this first and doing the email notification later pose a significant security risk?

CrystalPea:
@Wynndow @galund We updated this pull request with email alert functionality now. Let us know if you think it might be ready for merging ☺️

app/main/forms/auth_forms.py
```diff
@@ -44,14 +44,23 @@ class ChangePasswordForm(FlaskForm):
...
+        'Confirm new password', id="input_confirm_password",
         validators=[
             DataRequired(message="Please confirm your new password"),
             EqualTo('password', message="The passwords you entered do not match")
         ]
     )
 
 
+class ChangeOldPasswordForm(ChangePasswordForm):
+    old_password = PasswordField(
```
galund: Per @Wynndow's comments, I think I agree that this form isn't named very clearly - it could either do with a different name or with a very brief docblock explaining what it's for.

Why doesn't the ChangePasswordForm always ask for the old password? Maybe it needs renaming, if it's actually only used when the account is first created? Some confusion here anywhere that could be to do with naming things better.

app/main/views/reset_password.py
```diff
@@ -139,3 +141,87 @@ def update_password(token):
...
+                        current_app.logger.error(
+                            "{code}: Password change alert email for email_hash {email_hash} failed to send. "
+                            "Error: {error}",
+                            extra={
+                                'email_hash': hash_string(current_user.email_address),
+                                'error': str(e),
+                                'code': 'login.password-change-alert-email.notify-error'
+                            }
+                        )
+                        abort(503, response="Failed to send password change alert.")
```
Wynndow: If we abort here, the user will just see a 500, and won't know that their password has actually been changed. Does that matter?

app/templates/auth/change-password.html
```diff
@@ -0,0 +1,122 @@
...
+        {% for field_name, field_errors in form.errors|dictsort|reverse if field_errors %}
+          {% for error in field_errors %}
+            <li><a href="#{{ form[field_name].name }}" class="validation-masthead-link">{{ form[field_name].label.text }}</a></li>
+          {% endfor %}
+        {% endfor %}
+        </ul>
+    </div>
+{% endif %}
+
+<header class="page-heading">
```
Wynndow: Could you use a page heading from the tookit? http://alphagov.github.io/digitalmarketplace-frontend-toolkit/page-heading.html

app/templates/auth/change-password.html
```diff
@@ -0,0 +1,122 @@
...
+</header>
+
+<form autocomplete="off" action="{{ url_for('.change_password') }}" method="POST">
+
+    <div class="grid-row">
+        <div class="column-two-thirds">
+            {{ form.hidden_tag() }}
+
+
+            {% if form.old_password.errors %}
```
Wynndow: Could we use the toolkit forms stuff for these forms? http://alphagov.github.io/digitalmarketplace-frontend-toolkit/forms/textbox.html

app/main/views/reset_password.py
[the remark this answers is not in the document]
katstevens: This `try..except` was copied from the password reset view. I think the idea was that we wanted it to fail noisily for our benefit rather than for the user's, so we know that Notify is having problems? Agreed that it's not very useful for the user to get a 500 though.

app/templates/auth/change-password.html
katstevens: It's all copied and pasted from the reset_password template 😸 Happy to do a bit of refactoring so we're consistent across the User FE, but not sure if this branch is the place to do it?

tests/main/views/test_reset_password.py
```diff
@@ -251,3 +253,191 @@ def test_should_be_an_error_if_send_email_fails(self, send_email):
...
+        super().teardown_method(method)
+
+    @pytest.mark.parametrize(
+        'user_role, redirect_url',
+        [('buyer', '/buyers'), ('supplier', '/suppliers'), ('admin', '/admin')]
+    )
+    def test_change_password_page_displays_correctly(self, user_role, redirect_url):
+        if user_role == 'buyer':
+            self.login_as_buyer()
+        elif user_role.startswith('admin'):
```
Wynndow: Why not just use user_role == 'admin'?

tests/main/views/test_reset_password.py
```diff
@@ -251,3 +253,191 @@ def test_should_be_an_error_if_send_email_fails(self, send_email):
...
+        )
+        assert response.status_code == 302
+        assert response.location == 'http://localhost{}'.format(redirect_url)
+
+        self.data_api_client.update_user_password.assert_called_once_with(123, '0987654321', updater=user_email)
+        self.assert_flashes("You have successfully changed your password.")
+
+        send_email.assert_called_once_with(
+            user_email,
+            template_id='1c4c0562-44aa-4ae4-ba61-e17c544df535',
```
Wynndow: Could this id be imported from the config in case it changes?

tests/main/views/test_reset_password.py
katstevens: Because there are now five (count them!) admin roles - `admin-ccs-sourcing`, `admin-manager` and so on.

tests/main/views/test_reset_password.py
```diff
@@ -251,3 +253,191 @@ def test_should_be_an_error_if_send_email_fails(self, send_email):
...
+        self.data_api_client.authenticate_user.return_value = None
+        response = self.client.post(
+            '/user/change-password',
+            data={
+                'old_password': '0987654321',
+                'password': '1234567890',
+                'confirm_password': '1234567890'
+            }
+        )
+        assert self.strip_all_whitespace("Make sure you’ve entered the right password.") \
```
Wynndow: Could move this validation message to a constant like the others.

app/main/views/reset_password.py
Wynndow: Ah sorry, didn't see George's comment.

app/templates/auth/change-password.html
Wynndow: Maybe its a tech debt ticket.

app/main/forms/auth_forms.py
katstevens: Yeah, it's a bit confusing. We were reluctant to change the password reset flow too much but `ChangePasswordForm` is no longer the right name for that form! I'll take a look at that now.

tests/main/views/test_reset_password.py
Wynndow: But isn't `user-role` just coming in from the paramterization which only has `admin`?

tests/main/views/test_reset_password.py
katstevens: Yep - I'll do the same for the password reset while I'm at it.

tests/main/views/test_reset_password.py
katstevens: Haha you're right, sorry. I thought this was from the helper function we added (https://github.com/alphagov/digitalmarketplace-user-frontend/blob/master/app/main/helpers/login_helpers.py#L24). Yep this can be changed 😸

galund: Perhaps if there are two things here then we need `user_role` and `user_role_type` or _something_

katstevens: Nah I think it was just a poor choice of parameter - it should be `admin-ccs-sourcing` or similar, as we no longer use a straight `admin` role. I'll update it to be 💯clear.

katstevens: My mistake - we do still have `admin` as a valid role. Going to keep as is for now.

katstevens:
OK, I've pushed a commit that changes the flow so the user sees the 'success' message even if the alert email fails. Instead of raising a 5xx error, we just log an error message.

After trying and failing to find a sensible way to assert that we've done that logging, I settled for moving the logging itself to a new helper. This means we can patch `current_app.logger` without affecting the test client's request. It moves away from what I've been trying to do regarding helper functions, which is to keep Flask objects like `current_app`, `current_user` etc in the app views only, so we can better keep track of what app context is being applied where. If anyone can see a another way round it I'd be happy to change this.

katstevens:
@Wynndow @galund I think all the comments have been addressed now.

PR #62, opened 2018-05-15
Change password hint text wording

katstevens:
Trello: https://trello.com/c/TieWrBG5/44-password-reset-link-from-logged-in-session

Following a requested content change on the password change form to remove hint text - previously was `Repeat password used above`. I've updated the password reset form where this was copied from as well: 
change-password-no-hint-text


katstevens:
The functional tests don't check the hint text (neither do the unit tests!). Checked with @JaneEastwood and @constancecerf that the hint text should be removed entirely on both password change and password reset forms.

PR #64, opened 2018-05-29
Consolidate validation mastheads

samuelhwilliams:
## Summary
Use validation masthead from the frontend-toolkit consistently.

Also converts most pages over to using the base page from the frontend-toolkit.

 ## Ticket
https://trello.com/c/RSYEM3FL/448

 ## Requires
https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/440

### DIFF
--- a/app/main/forms/auth_forms.py
+++ b/app/main/forms/auth_forms.py
@@ -4,6 +4,12 @@ from wtforms.validators import DataRequired, EqualTo, Length, Regexp
 from dmutils.forms import StripWhitespaceStringField, StringField
 
 
+EMAIL_LOGIN_HINT = "Enter the email address you used to register with the Digital Marketplace"
+PASSWORD_HINT = "Must be between 10 and 50 characters"
+PHONE_NUMBER_HINT = "If there are any urgent problems with your requirements, we need your phone number so the " \
+                    "support team can help you fix them quickly."
+
+
 class LoginForm(FlaskForm):
     email_address = StripWhitespaceStringField(
         'Email address', id="input_email_address",
@@ -19,6 +25,10 @@ class LoginForm(FlaskForm):
             DataRequired(message="You must provide your password")
         ]
     )
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.email_address.hint = EMAIL_LOGIN_HINT
 
 
 class EmailAddressForm(FlaskForm):
@@ -56,6 +66,10 @@ class PasswordChangeForm(FlaskForm):
             EqualTo('password', message="The passwords you entered do not match")
         ]
     )
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.password.hint = PASSWORD_HINT
 
 
 class PasswordResetForm(PasswordChangeForm):
@@ -98,3 +112,8 @@ class CreateUserForm(FlaskForm):
                    )
         ]
     )
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.phone_number.hint = PHONE_NUMBER_HINT
+        self.password.hint = PASSWORD_HINT
--- a/app/main/forms/auth_forms.py
+++ b/app/main/forms/auth_forms.py
@@ -1,7 +1,13 @@
+from flask import current_app
+from flask_login import current_user
 from flask_wtf import FlaskForm
 from wtforms import PasswordField
-from wtforms.validators import DataRequired, EqualTo, Length, Regexp
+from wtforms.validators import DataRequired, EqualTo, Length, Regexp, ValidationError
+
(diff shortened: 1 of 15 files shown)


app/main/views/auth.py
```diff
@@ -32,9 +33,14 @@ def render_login():
     next_url = request.args.get('next')
     if current_user.is_authenticated() and not get_flashed_messages():
         return redirect_logged_in_user(next_url)
+
+    form = LoginForm()
```
CrystalPea: just out of curiosity, what is the reason for first defining the form and only then passing it in?

samuelhwilliams: I need to reference the form in two places - on the line below to get errors from it, and then again to pass it into `render_template` :)

CrystalPea: Ahh, how did I not see that, cheers! :)

app/templates/auth/change-password.html
```diff
@@ -38,60 +33,37 @@
...
-            {% if form.confirm_password.errors %}
-                </div>
-            {% endif %}
+          
+            {% with question = form.old_password.label,
+                    name = "old_password",
+                    error = errors.get('old_password', {}).get('message'),
+                    value = form.old_password.data,
+                    type = "password",
+                    autocomplete = "off"
```
katstevens: Do we need to state `autocomplete = "off"` if this is going to be the default for password type fields?

samuelhwilliams: It will be the default, but I also felt I wanted to be explicit in these forms.

app/main/views/reset_password.py
```diff
@@ -150,88 +150,63 @@ def update_password(token):
...
-            # Make sure old password is valid
-            user_json = data_api_client.authenticate_user(
-                current_user.email_address,
-                form.old_password.data)
-            if user_json:
-                response = data_api_client.update_user_password(
-                    current_user.id,
-                    form.password.data,
-                    updater=current_user.email_address
+    if form.validate_on_submit():
```
katstevens: Looks good, however I'd add a comment here to mention that the `old_password` check is now happening during `validate_on_submit` - I can foresee a dev looking at this code in the future (i.e. forgetful me, in 6 months' time) and deciding to be 'helpful' by adding it back in...

samuelhwilliams: Done 👍

app/main/forms/auth_forms.py
```diff
@@ -1,8 +1,20 @@
...
+from flask_login import current_user
 from flask_wtf import FlaskForm
 from wtforms import PasswordField
-from wtforms.validators import DataRequired, EqualTo, Length, Regexp
+from wtforms.validators import DataRequired, EqualTo, Length, Regexp, ValidationError
+
+from dmutils.email.helpers import hash_string
 from dmutils.forms import StripWhitespaceStringField, StringField
 
+from app import data_api_client
```
katstevens: 😿 I've been trying to avoid importing the `data_api_client` outside the apps' `views.py` (so we can keep track of the scope of our globals), but I can't see a better way of doing it for the validators. We do the same thing for  one of the Admin form validators (https://github.com/alphagov/digitalmarketplace-admin-frontend/blob/master/app/main/forms.py#L32). My preference for keeping the validation-related code here in `forms.py` is stronger than my pref for avoiding this import pattern!

app/main/views/reset_password.py
```diff
@@ -140,98 +143,71 @@ def update_password(token):
...
-
-                    flash(PASSWORD_UPDATED_MESSAGE)
-                else:
-                    flash(PASSWORD_NOT_UPDATED_MESSAGE, 'error')
-                return redirect(dashboard_url)
-            else:
-                current_app.logger.info(
-                    "change_password.fail: failed to authenticate user {email_hash}",
-                    extra={'email_hash': hash_string(current_user.email_address)})
+    # Checking that the old password is correct is done as a validator on the form.
```
katstevens: 👍

tests/main/views/test_reset_password.py
```diff
@@ -298,8 +304,22 @@ def setup_method(self, method):
         )
         self.data_api_client = self.data_api_client_patch.start()
 
+        self.auth_forms_data_api_client_patch = mock.patch(
+            'app.main.forms.auth_forms.data_api_client', **data_api_client_config,
+            **auth_forms_extra_data_api_client_config
+        )
+        self.auth_forms_data_api_client = self.auth_forms_data_api_client_patch.start()
+
+        self.auth_forms_current_app_patch = mock.patch('app.main.forms.auth_forms.current_app')
```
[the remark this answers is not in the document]
samuelhwilliams: Nice 👍

app/main/forms/auth_forms.py
samuelhwilliams: It is sad indeed to use the globals, but I agree that this is probably better than the alternative :(

PR #82, opened 2018-08-15
User FE error handling

katstevens:
Trello: https://trello.com/c/3iwy5gAy/115-pull-in-error-handling-to-fe-apps

- Pull in utils/toolkit, and update the form fields as per v42 changes https://github.com/alphagov/digitalmarketplace-utils/blob/master/CHANGELOG.md Some of the updated fields require `autocomplete = off`, so we can't use the contextless `{{ form.email_address }}` in templates yet.
- Remove old error templates
- In order to display custom error messages (e.g. 'The link you used to create an account is not valid....'), render the toolkit error templates directly from the views where needed.


app/main/forms/auth_forms.py
```diff
@@ -18,8 +18,9 @@
 
 
 class LoginForm(FlaskForm):
-    email_address = StripWhitespaceStringField(
+    email_address = DMStripWhitespaceStringField(
         'Email address', id="input_email_address",
+        hint=EMAIL_LOGIN_HINT,
```
lfdebrux: 👍

app/main/views/create_user.py
```diff
@@ -30,7 +30,10 @@ def create_user(encoded_token):
             "createuser.token_invalid: {encoded_token}",
             extra={'encoded_token': encoded_token}
         )
-        abort(400, INVALID_TOKEN_MESSAGE)
+        return render_template(
+            "toolkit/errors/400.html",
+            error_message=INVALID_TOKEN_MESSAGE,
+        ), 400
```
lfdebrux: Is there any reason we can't use render_error_page here?

katstevens: Kind of. Our `400` handler doesn't pass the error_message through to the template as it goes through `csrf_handler` first. Once we upgrade Flask past 0.11 we won't need to do that anymore.

lfdebrux: Okay, I guess we can put a TODO or record it as tech debt and come back to it later

PR #93, opened 2018-11-02
password forms: disallow blacklisted passwords

risicle:
https://trello.com/c/1dr27QpR/207-password-complexity-implement-check-display-guidance-if-weak

I've gone with the 10k password list, which should give us ~350 passwords >=10 chars.

I could have made this a hyper-generic "blacklist" wtforms validator but then I would have spend ages deciding what exactly to do about class-wide or instance-wide caching of the set etc... and right at the core of it there's still a big blatant "password"-specific pragma where we filter out terms >= `MIN_PASSWORD_LENGTH`. So, enough mental backflips - bedone with it.

While I was at it I found the existing tests covering these views to be a bit underwhelming so have improved their assertiveness.


app/data/password_blacklist/site_specific.txt
```diff
@@ -0,0 +1,17 @@
...
+gcloud
+g-cloud
+g_cloud
+g cloud
+procurement
+opportunity
+opportunities
+requirement
+requirements
+commercial
```
katstevens: Do we want to bother with l33t versions of these too - e.g. `d1g1t4lm4rk3tpl4c3`? Or are those strong enough 😸

app/main/forms/auth_forms.py
```diff
@@ -126,10 +177,12 @@ class CreateUserForm(FlaskForm):
...
         validators=[
             DataRequired(message="You must enter a password"),
-            Length(min=10,
-                   max=50,
-                   message="Passwords must be between 10 and 50 characters"
-                   )
+            Length(
+                min=PASSWORD_MIN_LENGTH,
+                max=50,
+                message=f"Passwords must be between {PASSWORD_MIN_LENGTH} and 50 characters",
```
lfdebrux: Can we use `PASSWORD_HINT` here?

app/main/forms/auth_forms.py
```diff
@@ -73,10 +122,12 @@ class PasswordChangeForm(FlaskForm):
...
         validators=[
             DataRequired(message="You must enter a new password"),
-            Length(min=10,
-                   max=50,
-                   message="Passwords must be between 10 and 50 characters"
-                   )
+            Length(
+                min=PASSWORD_MIN_LENGTH,
+                max=50,
+                message=f"Passwords must be between {PASSWORD_MIN_LENGTH} and 50 characters",
```
lfdebrux: Can we use `PASSWORD_HINT` here?

app/main/forms/auth_forms.py
```diff
@@ -73,10 +122,12 @@ class PasswordChangeForm(FlaskForm):
...
+            ),
+            NotInPasswordBlacklist(message="Your password must be hard to guess"),
```
lfdebrux: Can we have a `PASSWORD_IN_BLACKLIST_HINT` class variable for this text?

app/main/forms/auth_forms.py
```diff
@@ -126,10 +177,12 @@ class CreateUserForm(FlaskForm):
...
```
lfdebrux: Can we have a `PASSWORD_IN_BLACKLIST_HINT` class variable for this text?

app/main/forms/auth_forms.py
```diff
@@ -17,6 +20,52 @@
                     "support team can help you fix them quickly."
 
 
+PASSWORD_MIN_LENGTH = 10
```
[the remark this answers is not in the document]
risicle: Um, don't see why not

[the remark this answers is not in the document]
risicle: but I'll call it something more specific than `PASSWORD_HINT`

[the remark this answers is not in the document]
risicle: ...*oh* I see how it uses `PASSWORD_HINT`...

PR #103, opened 2019-02-18
Update requirements-dev

lfdebrux:
Bumps outdated dependencies in `requirements-dev.txt`. This is instead of merging #102.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -10,7 +10,7 @@ ignore = D203,W503
 max-complexity = 7
 max-line-length = 120
 per-file-ignores =
-    **/__init__.py : F401
-    app/main/__init__.py : E402
-    app/status/__init__.py : E402
+    app/__init__.py : F401
+    app/main/__init__.py : E402, F401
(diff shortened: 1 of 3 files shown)


.flake8
```diff
@@ -10,7 +10,7 @@ ignore = D203,W503
 max-complexity = 7
 max-line-length = 120
 per-file-ignores =
-    **/__init__.py : F401
-    app/main/__init__.py : E402
-    app/status/__init__.py : E402
+    app/__init__.py : F401
```
jonodrew: This works as `**/_init__.py` and `app/__init__.py` - can you say why you changed it?

lfdebrux: `**/__init__.py` in the old plugin would get all files called `__init__.py`, whereas now it would only get `app/__init__.py`.

So leaving it in is just leaving around a bit of code that doesn't make sense anymore.

jonodrew: 👍

PR #107, opened 2019-03-11
password reset mechanism: handle inactive users more gracefully

risicle:
https://trello.com/c/ARkdAn0z

Depends on https://github.com/alphagov/digitalmarketplace-utils/pull/500

At token generation time, if the user turns out to be inactive, instead of sending a password reset token we send the user a message explaining that their account is disabled and what they can do about it.

~Still to do: decide on the content of this message, write the Notify template and paste its template id in to the `config`.~

~Edit: oh and write some functional tests.~

### DIFF
--- a/app/main/views/reset_password.py
+++ b/app/main/views/reset_password.py
@@ -50,47 +50,72 @@ def send_reset_password_email():
         user_json = data_api_client.get_user(email_address=email_address)
 
         if user_json is not None:
-
             user = User.from_json(user_json)
-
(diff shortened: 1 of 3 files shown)


katstevens:
Yep, looks good. Will thumbs up once all the bits and pieces are ready.

lfdebrux:
Marked as WIP to make the seal happy

risicle:
Ok, removing WIP marker...

PR #109, opened 2019-03-18
Update digitalmarketplace-utils

lfdebrux:


requirements.txt
```diff
@@ -43,9 +43,9 @@ PyJWT==1.7.1
 python-dateutil==2.8.0
 python-json-logger==0.1.10
 pytz==2018.9
-PyYAML==3.11
+PyYAML==3.13
 requests==2.21.0
-s3transfer==0.1.13
+s3transfer==0.2.0
```
benvand: https://pypi.org/project/s3transfer/

benvand: > This project is not currently GA. If you are planning to use this code in production, make sure to lock to a minor version as interfaces may break from minor version to minor version. For a basic, stable interface of s3transfer, try the interfaces exposed in boto3

benvand: It's boto3 bringing this in 🤦‍♂️

lfdebrux: Oh fun times

lfdebrux: Looks like we don't actually use s3transfer anywhere, just boto3

PR #114, opened 2019-04-24
Bump digitalmarketplace-utils dependency to v48.2.0

risicle:
https://trello.com/c/9l0YBPl5

Also bump `digitalmarketplace-content-loader` to `5.2.1` and other version bumps that come with the re-freeze.

This includes werkzeug 0.15, which requires a couple of minor changes to our tests that depend on the default behaviour of werkzeug/flask, which has changed but not in a way that I think is particularly important to us.

To make it clear that one of the tests is simply being affected by the change in default base path for absolutizing url redirects (and has nothing to do with `@` handling) I added another test which simply exhibits the path absolutizing on its own.

The "default default" redirect code has also changed to 308, which I've reflected in a single test change.

### DIFF
--- a/tests/main/views/test_auth.py
+++ b/tests/main/views/test_auth.py
@@ -83,13 +83,13 @@ class TestLogin(BaseApplicationTest):
         self.login_as_supplier()
         res = self.client.get("/user/login?next=dolphinsbarn")
         assert res.status_code == 302
-        assert res.location == 'http://localhost/dolphinsbarn'
+        assert res.location == 'http://localhost/user/dolphinsbarn'
 
     def test_should_redirect_to_next_url_for_simple_auth_uri(self):
         self.login_as_supplier()
         res = self.client.get("/user/login?next=@example.com")
         assert res.status_code == 302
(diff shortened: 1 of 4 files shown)


tests/main/views/test_auth.py
```diff
@@ -79,11 +79,17 @@ def test_should_redirect_to_dashboard_if_next_url_is_not_http(self):
...
+        res = self.client.get("/user/login?next=dolphinsbarn")
+        assert res.status_code == 302
+        assert res.location == 'http://localhost/user/dolphinsbarn'
+
     def test_should_redirect_to_next_url_for_simple_auth_uri(self):
         self.login_as_supplier()
         res = self.client.get("/user/login?next=@example.com")
         assert res.status_code == 302
-        assert res.location == 'http://localhost/@example.com'
+        assert res.location == 'http://localhost/user/@example.com'
```
katstevens: Functional tests all ok with this?

risicle: :thinking: good question

risicle: It seems so, yes.

PR #119, opened 2019-05-14
495-add-metrics-to-briefs-user-and-admin-frontends

benvand:
* Add dependencies
* Add metrics
* Add tests

https://trello.com/c/Fqq8RCEb/495-add-metrics-to-briefs-user-and-admin-frontends

### DIFF
--- a/app/metrics.py
+++ b/app/metrics.py
@@ -1,0 +1,9 @@
+from flask import Blueprint
+from dmutils.metrics import DMGDSMetrics
+
+
+metrics = Blueprint('metrics', __name__)
+
+gds_metrics = DMGDSMetrics()
+
+metrics.add_url_rule(gds_metrics.metrics_path, 'metrics', gds_metrics.metrics_endpoint)
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -19,12 +19,16 @@ class BaseApplicationTest(object):
         # as the DataAPIClient instance has already been created; nor can we temporarily replace app.data_api_client
         # with a mock, because then the shared instance won't have been configured (done in create_app). Instead,
         # just mock the one function that would make an API call in this case.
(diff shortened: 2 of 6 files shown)


app/metrics.py
```diff
@@ -0,0 +1,31 @@
+from flask import Blueprint
+from flask.signals import got_request_exception, request_finished
+
+from gds_metrics import GDSMetrics
+
+
+metrics = Blueprint('metrics', __name__)
+
+
+class DMGDSMetrics(GDSMetrics):
```
katstevens: Didn't we move this to utils? Am I getting confused?

benvand: Good shout, copied the wrong file

tests/helpers.py
```diff
@@ -19,12 +19,16 @@ def setup_method(self, method):
         # as the DataAPIClient instance has already been created; nor can we temporarily replace app.data_api_client
         # with a mock, because then the shared instance won't have been configured (done in create_app). Instead,
         # just mock the one function that would make an API call in this case.
+        self.app_env_var_mock = patch.dict('gds_metrics.os.environ', {'PROMETHEUS_METRICS_PATH': '/_metrics'})
```
lfdebrux: I'm confused, should this be `/user/_metrics` as in the tests below?

benvand: No, we register the metrics blueprint prefixed at `'/user'` in the app setup here:
https://github.com/alphagov/digitalmarketplace-user-frontend/pull/119/files#diff-1f9f743421417fb2794c88447083a031R33

The `gds_metrics` library defines the url suffix using:
`gds_metrics.os.environ['PROMETHEUS_METRICS_PATH']`
here:
https://github.com/alphagov/gds_metrics_python/blob/master/gds_metrics/__init__.py#L30

The tests set up the app the same as it is in prod, we initialise the metrics which grabs `_metrics` from the `PROMETHEUS_METRICS_PATH` env var, then we prefix the metrics blueprint with `/user`

lfdebrux: Okay gotcha

PR #133, opened 2019-06-10
Remove report a problem form

lfdebrux:
https://trello.com/c/z63XpU1l/548-remove-is-there-anything-wrong-with-this-page-form

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "jquery": "1.12.0",
     "hogan.js": "3.0.2",
     "jquery-details": "https://github.com/mathiasbynens/jquery-details/archive/v0.1.0.tar.gz",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -19,7 +19,7 @@
     "jquery": "1.12.0",
     "hogan.js": "3.0.2",
     "jquery-details": "https://github.com/mathiasbynens/jquery-details/archive/v0.1.0.tar.gz",
-    "digitalmarketplace-frontend-toolkit": "https://github.com/alphagov/digitalmarketplace-frontend-toolkit.git#v31.8.0",
+    "digitalmarketplace-frontend-toolkit": "https://github.com/alphagov/digitalmarketplace-frontend-toolkit.git#v33.0.0",
```
katstevens: This should be OK with the node/gulp bump at v32, right?

lfdebrux: We haven't had any issues with any of the other apps, and the tests passed locally, so I think so?

PR #138, opened 2019-07-16
Disallow admin-manager from using password reset

lfdebrux:
Trello: https://trello.com/c/lkytm0e1/604-disable-the-password-reset-mechanism-for-the-admin-manager

### DIFF
--- a/tests/main/views/test_reset_password.py
+++ b/tests/main/views/test_reset_password.py
@@ -325,6 +325,22 @@ class TestResetPassword(BaseApplicationTest):
             reference='reset-password-inactive-8yc90Y2VvBnVHT5jVuSmeebxOCRJcnKicOe7VAsKu50=',
         )
 
+    @mock.patch("app.main.views.reset_password.DMNotifyClient.send_email", autospec=True)
+    def test_admin_manager_does_not_get_reset_email(self, send_email):
+        self.data_api_client.get_user.return_value = self.user(
+            123, "email@email.com", name="Eve", role="admin-manager",
+            supplier_id=None, supplier_name=None,
+        )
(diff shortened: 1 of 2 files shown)


tests/main/views/test_reset_password.py
```diff
@@ -324,6 +324,22 @@ def test_inactive_user_attempts_password_reset(self, send_email):
...
+        self.data_api_client.get_user.return_value = self.user(
+            123, "email@email.com", name="Eve", role="admin-manager",
+            supplier_id=None, supplier_name=None,
+        )
+        res = self.client.post(
+            "/user/reset-password",
+            data={"email_address": "email@email.com"}
+        )
+
+        # to prevent revealing email addresses for admins we still show the usual result
```
risicle: Ahhhhhhhhhh :brain:

lfdebrux:
> As discussed offline: let's add a `TODO: handle genuine reset requests for admin managers` or similar.

I've put together a new Trello ticket: https://trello.com/c/i5HObZsb/959-how-do-we-handle-password-reset-requests-for-admin-manager-accounts

I'll hold off on merging this until https://github.com/alphagov/digitalmarketplace-user-frontend/pull/139 is merged so I can then include the fix you mentioned @katstevens

risicle:
Indeed I think this `format()` call is one of those places where `Markup` will be clever and escape the new text as it's inserted.

lfdebrux:
> Indeed I think this `format()` call is one of those places where `Markup` will be clever and escape the new text as it's inserted.

Are you talking about #139? I think you might have posted on the wrong PR.

risicle:
God damnit wrong window

PR #139, opened 2019-07-16
Move support email to app config

katstevens:
https://trello.com/c/3om47rfx/601-update-remaining-instances-of-enquiriesdigitalmarketplaceservicegovuk-to-the-new-support-address

Centralises the support email rather than hardcoding it in:
- create user templates/flash messages
- reset password templates/flash messages (see below)

reset-password-help-emai

### DIFF
--- a/config.py
+++ b/config.py
@@ -30,6 +30,7 @@ class Config(object):
         "change_password_alert": "1c4c0562-44aa-4ae4-ba61-e17c544df535",
         "reset_password_inactive": "6c522c78-e4d2-488f-aa5f-6f42401ef2c5",
     }
+    SUPPORT_EMAIL_ADDRESS = "cloud_digital@crowncommercial.gov.uk"
 
     DEBUG = False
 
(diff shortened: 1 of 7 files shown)


config.py
```diff
@@ -30,6 +30,7 @@ class Config(object):
         "change_password_alert": "1c4c0562-44aa-4ae4-ba61-e17c544df535",
         "reset_password_inactive": "6c522c78-e4d2-488f-aa5f-6f42401ef2c5",
     }
+    SUPPORT_EMAIL_ADDRESS = "cloud_digital@crowncommercial.gov.uk"
```
risicle: Underscore? Really?

lfdebrux:
Can we use f-strings instead of expecting the dev to call `format` on the templates?

config.py
katstevens: Not our decision...

katstevens:
@lfdebrux I'm not sure whether the F-strings will play nicely with the `Markup` objects we're using.

katstevens:
@risicle I think the Supplier FE is the only app where the functional tests are looking for the `enquiries@` email.

PR #141, opened 2019-07-25
send_reset_password_email: fix check of user.role being "admin-manager" 

risicle:
https://trello.com/c/o7nrGSfH

Without a comma this is just interpreted as a set of ignorable brackets.

Also improved the tests around this. This is a perfect example of why "one test, one feature" doesn't cut it.

### DIFF
--- a/tests/main/views/test_reset_password.py
+++ b/tests/main/views/test_reset_password.py
@@ -262,8 +262,20 @@ class TestResetPassword(BaseApplicationTest):
         assert reset_password.EXPIRED_PASSWORD_RESET_TOKEN_MESSAGE in error_elements[0].text_content()
         assert self.data_api_client.update_user_password.called is False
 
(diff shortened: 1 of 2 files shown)


tests/main/views/test_reset_password.py
```diff
@@ -262,8 +262,16 @@ def test_token_created_before_last_updated_password_cannot_be_used(self):
         assert reset_password.EXPIRED_PASSWORD_RESET_TOKEN_MESSAGE in error_elements[0].text_content()
         assert self.data_api_client.update_user_password.called is False
 
+    @pytest.mark.parametrize("user_role", (
+        "admin",
```
katstevens: Might as well add the other admin roles in as well (`admin-ccs-category` etc)?

risicle: :heavy_check_mark:

PR #144, opened 2019-08-05
Force pyup to ignore flask updates

jonodrew:
We're not certain that the new version of Flask is stable, so we're going to ignore it for a couple of weeks until they've had a chance to iron out the bugs


risicle:
You'll need to "freeze" this one too I guess (or just copy the comment to `requirements.txt`)

PR #147, opened 2019-09-03
Update NodeJS from version 8 to version 10

lfdebrux:
https://trello.com/c/YknBKmMU/261-update-node-from-8-to-10


package.json
```diff
@@ -3,24 +3,26 @@
   "description": "User application for the Digital Marketplace",
   "version": "1.0.0",
   "private": true,
-  "engine": "node >= 6.10.3 < 7.0.0",
+  "engines": {
+    "node": "~ 10.16.3"
```
risicle: Can we just use `^` here instead? It's not the job of this file to do the actual installing is it?

risicle: (in fact it would have to be `^ 10.15` - nix stable has a patched 10.15 version)

risicle:
Pushed a nix fixup.

lfdebrux:
> Pushed a nix fixup.

I will make sure this patch gets applied to all apps now

risicle:
Oh don't worry i've got it...

PR #155, opened 2019-09-13
Allow visiting `/users/logout` with a GET request

lfdebrux:
## Summary

We might want to be able to create a link to logout a user using the `<a>` HTML tag. I've created this PR as a strawman that we can discuss.

## Details

In our current frontend apps we use a form to ask the browser to make a POST request to the logout url. This has some advantages on the backend (it's clear that the request is destructive, we can include a CSRF token), but on the frontend this causes issues with the log out button.

In the current admin frontend the focus state of a form differs visually from the focus state of a link.

In the current admin frontend the focus state of a form differs visually from the focus state of a link

With the GOV.UK Frontend, this problem is exacerbated as currently the header component does not support arbitrary HTML, and the form styles are designed for multiple inputs, not just one button.

In the admin frontend with the GOV.UK Frontend header focusing on the log out button using the keyboard requires pressing tab multiple times to move through the form

These issues with the focus state might not be easy to fix if we continue to use a form for the log out button; ideally the button should be a simple link. However, a HTML link can only make a GET request (unless JavaScript is used), so we have to allow GET requests to the `logout` route.

## Possible issues

There is some suggestion that having a link to a log out endpoint might cause users to be logged out by aggressive browser pre-caching (see Stack Overflow issues [1](https://stackoverflow.com/a/14587231) [2](https://stackoverflow.com/a/15098437)), however I have tested this locally with the admin frontend and there did not seem to be any issue logging in and staying logged in.

### DIFF
--- a/app/main/views/auth.py
+++ b/app/main/views/auth.py
@@ -79,7 +79,12 @@ def process_login():
             next=next_url), 400
 
 
-@main.route('/logout', methods=["POST"])
+# We allow logging out via GET request so that we can have a simple link in the
+# site header. We would prefer to be able to have a degree of protection
+# against inadvertent/malicious logout requests without the user knowing, which
+# is why we have used POST previously, but our site header design needs work
+# before that can happen.
+@main.route('/logout', methods=["GET", "POST"])
 def logout():
     session.clear()
     logout_user()
--- a/tests/main/views/test_auth.py
+++ b/tests/main/views/test_auth.py
@@ -186,6 +186,14 @@ class TestLogin(BaseApplicationTest):
         with self.client.session_transaction() as session:
             assert session.get('company_name') is None
 
+    @mock.patch("app.main.views.auth.logout_user")
+    def test_visiting_logout_should_logout(self, logout_user):
+        self.login_as_supplier()
+        res = self.client.get('/user/logout')
+        assert logout_user.called
+        assert res.status_code == 302
+        assert res.location == 'http://localhost/user/login'
+
     def test_should_return_a_403_for_invalid_login(self):
         self.data_api_client.authenticate_user.return_value = None
 


lfdebrux:
Addendum: @aliuk2012 has raised this as an issue against the GOV.UK Frontend [3](https://github.com/alphagov/govuk-frontend/issues/1573), so this might not end up being an issue,  but I thought I would raise this as a PR anyway because it would be in my mind much simpler and nicer to just be able to create a link to logout.

katstevens:
Interesting. I wonder if we should be using a 'link-like' thing in the first place to signify an action. A button in the header is obviously not a goer - could we feasibly put it anywhere else?

risicle:
It's certainly not an uncommon pattern to "break the rules" by making the logout view a `GET` request.

But it's arguable that in our case it's not even breaking the rules because logging out is entirely client-side. And if the worst comes to the worst, we could always add a cache-busting random parameter to the logout url.

benvand:
Would be good to have a functional test explicitly testing this functionality.

benvand:
Can we WIP this PR?

lfdebrux:
I'm reopening this as it's blocking https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/560.

Regarding functional tests that look at the HTML, that won't be possible until user frontend is using digitalmarketplace-govuk-frontend, so I think we should merge this now and add tests later.

katstevens:
Pro: removing the blocker, ditching old code etc.  We're not changing any server side data so a POST could be seen as overkill

Con: [Some browsers can pre-fetch URLs on a page](https://en.wikipedia.org/wiki/Link_prefetching#Issues_and_criticisms), so a logged in user could be logged out by the browser assuming that GET request was 'safe'.

risicle:
> Some browsers can pre-fetch URLs on a page, so a logged in user could be logged out by the browser assuming that GET request was 'safe'.

Yup, this is true, but IRL like 90% of web apps do it this way, so it's probably safe.

katstevens:
> got to make sure pages like this are always set no-cache

If it's in the header, that's... every page?

risicle:
Yes we're fine for now, but _if_ we ever did decide to do something drastic and turn it on globally, we'd have to remember this sort of page.

lfdebrux:
Okay, I've got two approvals so I'm going to merge it.

If it doesn't work we're not wedded to this approach; the POST endpoint will still work, so I don't think merging this is too big a step.

risicle:
Just one point - are we able to protect the endpoint in some way that will prevent cross-site annoyances? If we just accept any old GET request some ass can do something like link to an "image" in some random other page that will cause someone to "log themselves out" on load.

lfdebrux:
> Just one point - are we able to protect the endpoint in some way that will prevent cross-site annoyances? If we just accept any old GET request some ass can do something like link to an "image" in some random other page that will cause someone to "log themselves out" on load.

That'd definitely a thing that's possible. I'm not sure if it's anything that we should be that worried about though.

Most websites leave themselves open to that sort of exploit, I suspect because they feel it is not worth the effort of mitigating; cf https://www.quora.com/How-do-websites-like-Super-Logout-work

risicle:
Quite possible. An "easy" step would be to reject requests that don't have a referrer originating from our site. Or rejecting requests that don't `Accept` html.

lfdebrux:
> Quite possible. An "easy" step would be to reject requests that don't have a referrer originating from our site. Or rejecting requests that don't `Accept` html.

I'm happy for you to append a commit doing that.

risicle:
Meh if nobody else thinks it's something to worry about I'm ok with it.

katstevens:
I'm ok with this being merged to unblock the rest of the FE work, given that the risks of unexpected logout are low. But I'm uncomfortable with us endorsing this pattern in the long term. I'd rather move the Logout function out of the header entirely and/or make it a proper button. Obviously that would need some design input, which we don't have right now, and would cause more immediately confusion for our users. Could we at least add a comment to the code saying why it allows both GET and POST?

lfdebrux:
> I'm ok with this being merged to unblock the rest of the FE work, given that the risks of unexpected logout are low. But I'm uncomfortable with us endorsing this pattern in the long term. I'd rather move the Logout function out of the header entirely and/or make it a proper button. Obviously that would need some design input, which we don't have right now, and would cause more immediately confusion for our users. Could we at least add a comment to the code saying why it allows both GET and POST?

Understood. I will add the comment.

Regarding moving the logout functionality, perhaps it would be worth us putting out a feeler on the GOV.UK Design System backlog here https://github.com/alphagov/govuk-design-system-backlog/issues/95?

PR #158, opened 2019-09-24
Update blacklisted passwords and tests

jonodrew:
The list of blacklisted passwords has been increased by a factor of 10 to 100,000. One of the tests used a password that was not in the old list but is in the new, so I've updated that test as well


katstevens:
Does the bigger file make any significant difference to the timings of the view response? Hopefully it's not x10...

PR #159, opened 2019-09-25
Remove password used in tests

jonodrew:
This password is used in some tests and consequently was causing failures.


risicle:
This is also my password for all my production accounts so it would be good to allow it here too.

PR #161, opened 2019-10-09
Password reset sends sandbox email for nonexistent accounts

katstevens:
Trello: https://trello.com/c/PTzPDEqq/1166-address-user-enumeration-forgot-pw-function-vulnerability

Addresses the scenario where an attacker could determine the existence of an account by looking at the response time from the password reset view. To mitigate this, the view should do roughly the same actions for a non-existent user as for an existing inactive user. Local timing testing is always going to be a bit unreliable in terms of network requests so I'd like to see how this does on Preview before releasing.

Agreed with the Notify team that the sandbox email address is suitable for this usecase.

I also tidied up the tests for this view, as it was hard to tell what paths had coverage.

### DIFF
--- a/app/main/views/reset_password.py
+++ b/app/main/views/reset_password.py
@@ -94,7 +94,7 @@ def send_reset_password_email():
                     ), 503
 
                 current_app.logger.info(
-                    "{code}: Sending password reset email for email_hash {email_hash}",
+                    "{code}: Sent password reset email for email_hash {email_hash}",
                     extra={
                         'email_hash': hash_string(user.email_address),
                         'code': 'login.reset-email.sent'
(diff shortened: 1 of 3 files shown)


app/main/views/reset_password.py
```diff
@@ -127,8 +127,29 @@ def send_reset_password_email():
...
+                    email_address,  # Hashed by the helper function
+                )
+                return render_template(
+                    'toolkit/errors/500.html',
+                    error_message="Failed to send password reset.",
+                ), 503
+
             current_app.logger.info(
-                "{code}: Password reset request for invalid email_hash {email_hash}",
+                "{code}: Sending password (non-)reset email for invalid user email_hash {email_hash}",
```
risicle: _Technically_ it's already been sent.

katstevens: Good point!

katstevens: On closer inspection, we say the same thing for the other scenarios (valid user, invalid user). The request has been fired off to Notify, but the email itself probably hasn't been sent yet 😸 Shall we keep it consistent?

risicle: :man_shrugging: I'd just change them all - there's never any point that we _know_ our email's been sent really - it could always be spooling in some midway mail server somewhere.

katstevens: Fair enough - fixed.

PR #164, opened 2019-10-17
Add govuk-frontend-jinja and digitalmarketplace-govuk-frontend

risicle:
Hopefully complying to https://trello.com/c/ThyPYCdU

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -20,9 +20,12 @@
 //= require ../../../node_modules/digitalmarketplace-frontend-toolkit/toolkit/javascripts/show-hide-content.js
 //= require ../../../node_modules/digitalmarketplace-frontend-toolkit/toolkit/javascripts/user-research-consent-banner.js
 //= require ../../../node_modules/digitalmarketplace-frontend-toolkit/toolkit/javascripts/report-a-problem.js
+//= require ../../../node_modules/digitalmarketplace-govuk-frontend/govuk-frontend/all.js
 //= require _analytics.js'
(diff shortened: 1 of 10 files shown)


app/assets/javascripts/application.js
```diff
@@ -20,10 +20,13 @@
 //= require ../../../node_modules/digitalmarketplace-frontend-toolkit/toolkit/javascripts/show-hide-content.js
 //= require ../../../node_modules/digitalmarketplace-frontend-toolkit/toolkit/javascripts/user-research-consent-banner.js
 //= require ../../../node_modules/digitalmarketplace-frontend-toolkit/toolkit/javascripts/report-a-problem.js
+//= require ../../../node_modules/digitalmarketplace-govuk-frontend/govuk-frontend/all.js
 //= require _analytics.js'
 //= require _onready.js'
 //= require _selection-buttons.js
 
+GOVUKFrontend.initAll()
```
aliuk2012: Could you add either a `;` at the end of this line or place it on its own line at the bottom of the page?

This is causing the js to fail because the browser thinkings `GOVUKFrontend.initAll` returns a function  (which it doesn't) as the next block is a function. eg. GOVUKFrontend.initAll()()

risicle: `wtf.js`

aliuk2012: # 🤷‍♂.js

aliuk2012: The reason it wasn't added in admin frontend is that admin frontend only imported scripts and had no other content after the imports.

PR #165, opened 2019-10-18
switch to using digitalmarketplace-govuk-frontend for breadcrumbs, standardize trailing breadcrumb element

risicle:
https://trello.com/c/vcjVUo8x

In places where we didn't previously have a trailing breadcrumb element representing the current page I added one, mostly with the exact heading of the page but in some places using a bit of artistic license to make it a bit simpler (see user research mailing list page).

Spectacle nS6501

### DIFF
--- a/app/templates/auth/request-password-reset.html
+++ b/app/templates/auth/request-password-reset.html
@@ -3,23 +3,19 @@
 {% block page_title %}Reset password – Digital Marketplace{% endblock %}
 
 {% block breadcrumb %}
-  {%
-    with items = [
-      {
-        "link": "/",
-        "label": "Digital Marketplace"
-      },
-      {
-        "link": "/user/login",
-        "label": "Login"
-      },
-      {
-        "label": "Reset password"
-      }
-    ]
-  %}
-    {% include "toolkit/breadcrumb.html" %}
-  {% endwith %}
+  {{ govukBreadcrumbs({"items": [
+    {
+      "text": "Digital Marketplace",
+      "href": "/",
+    },
+    {
+      "text": "Login",
+      "href": "/user/login",
+    },
+    {
+      "text": "Reset password",
+    },
+  ]}) }}
 {% endblock %}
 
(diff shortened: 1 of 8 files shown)


risicle:
Now changed to "User research mailing list preferences"

app/templates/auth/request-password-reset.html
```diff
@@ -3,23 +3,19 @@
...
-    {% include "toolkit/breadcrumb.html" %}
-  {% endwith %}
+  {{ govukBreadcrumbs({"items": [
+    {
+      "text": "Digital Marketplace",
+      "href": "/",
+    },
+    {
+      "text": "Login",
+      "href": "/user/login",
```
aliuk2012: As part of the password reset user journey why would I need to go back to the user login page? 
Would I be able to authenticate again? would it invalidate my existing session?

app/templates/auth/request-password-reset.html
```diff
@@ -3,23 +3,19 @@
...
+    },
+    {
+      "text": "Reset password",
```
aliuk2012: Could this be "Request password reset" or "Request new password"?

Looking at the breadcrumbs I thought it was part of the change password user journey. It was only after looking at the filename did I realise that this was meant for user who had forgotten their password.

risicle: @pocketstefan!!

pocketstefan:
@aliuk2012 @risicle Hey Al, I think it makes sense to keep it as is for now. What page does the current breadcrumb 'Reset password' clash with?

risicle:
@aliuk2012 what shall I do with this? I'd like to get on with the rest of the changes and this is a bit of a blocker...

app/templates/auth/request-password-reset.html
aliuk2012: After discussion with @pocketstefan, we agreed to keep this the same 👍

PR #166, opened 2019-10-22
switch headings to digitalmarketplace-govuk-frontend, add temp spacing override

risicle:
https://trello.com/c/NwJZfDhf


lfdebrux:
Nice one.

I think you missed a file though; `app/templates/auth/create-user-error.html`?

risicle:
Ah right yeah - I started with a simplistic grep on `toolkit/page-heading`

lfdebrux:
Cool.

Additionally we should now be able to remove this line @import "toolkit/_page-headings.scss";

PR #175, opened 2019-10-29
switch grid elements to digitalmarketplace-govuk-frontend's classes

risicle:
https://trello.com/c/pLpLtMx1

This also quietly slips in a fix for a mistake from a previous merge, leaving us with mismatched tags and broken layout on the user research consent page. Al and I also agreed that the content there should be turned into a _bulleted_ list.

Otherwise the grid changes seem to make little difference.

### DIFF
--- a/app/templates/auth/change-password.html
+++ b/app/templates/auth/change-password.html
@@ -25,8 +25,8 @@
 
 <form autocomplete="off" action="{{ url_for('.change_password') }}" method="POST">
 
-    <div class="grid-row">
-        <div class="column-two-thirds">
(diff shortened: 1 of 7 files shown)


lfdebrux:
> This also quietly slips in a fix for a mistake from a previous merge, leaving us with mismatched tags and broken layout on the user research consent page.

Interesting, good spot. We should add the user research consent page to the visual regression tests.

katstevens:
Is it me or does this text seem really big? (It's like this on production so not introduced by this PR...)
user-fe-text-size

PR #177, opened 2019-10-31
Fix indenting to 2 spaces

aliuk2012:


lfdebrux:
Oh and if we are fixing on 2 space indents we should communicate/document this somewhere

PR #186, opened 2019-11-11
Removes dmspeak import from application.scss

aliuk2012:
We are no longer using the dmspeak classes so we should remove the import.


katstevens:
Though saying that, it's referenced in `_error-pages.scss`?

aliuk2012:
@katstevens Good spot, We shouldn't have `_error_pages.scss` either now that we are using digitalmarketplace-govuk-frontend error page templates. I'll make another PR for that.

PR #190, opened 2019-11-18
Replace Validation component with GOV.UK Error Summary

aliuk2012:
Requires https://github.com/alphagov/digitalmarketplace-utils/pull/543 to be merged and for this repo to be updated to use the changes.

Ticket: https://trello.com/c/jy1qTEQN/132-replace-validation-banner-with-govuk-frontend-error-summary-component-in-user-frontend

### DIFF
--- a/app/templates/_base_page.html
+++ b/app/templates/_base_page.html
@@ -4,7 +4,7 @@
 
 {% from "components/breadcrumbs/macro.njk" import govukBreadcrumbs %}
 {% from "components/button/macro.njk" import govukButton %}
-
+{% from "components/error-summary/macro.njk" import govukErrorSummary %}
 
 {# TODO: Remove once we start removing govuk_template, GOV.UK Frontend template uses pageTitle #}
 {% block page_title %}{% block pageTitle %}{% endblock %}{% endblock %}
@@ -18,6 +18,14 @@
     {% endblock %}
     <main id="content" role="main">
       {% include "toolkit/flash_messages.html" %}
+      {% block errorSummary %}
+        {% if errors %}
+          {{ govukErrorSummary({
+          "titleText": "There is a problem",
+          "errorList": errors.values()
+        }) }}
+        {% endif %}
+      {% endblock %}
       {# TODO: Remove this block when we start removing govuk_template but be sure #}
       {#       To keep the inner block mainContent otherwise no content will be rendered #}
       {% block main_content %}
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -12,7 +12,6 @@
 //= require ../../../node_modules/digitalmarketplace-frontend-toolkit/toolkit/javascripts/list-entry.js
 //= require ../../../node_modules/digitalmarketplace-frontend-toolkit/toolkit/javascripts/clear-filters.js
 //= require ../../../node_modules/digitalmarketplace-frontend-toolkit/toolkit/javascripts/word-counter.js
-//= require ../../../node_modules/digitalmarketplace-frontend-toolkit/toolkit/javascripts/validation.js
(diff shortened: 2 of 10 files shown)


app/templates/_base_page.html
```diff
@@ -4,6 +4,26 @@
...
+{% macro dmErrorSummary(params) %}
+  {% if params.errors %}
+    {% set locals = namespace(errorList = []) %}
+    {% for error in params.errors.values() %}
+      {% set locals.errorList = locals.errorList + [{
+        'text': error.message,
+        'href': '#input-' + error.input_name
+      }] %}
+
+    {% endfor %}
```
aliuk2012: If this could be done in the backend and returned to the view as part of `errors` it would help simplify the code and we call `govukErrorSummary` like this:

```
{% if errors %}
  {{ govukErrorSummary({
       "title": "There is a problem",
       "errorList": errors.errorList})
  }}
{% endif %}
```

lfdebrux: Let's see what we can do!

lfdebrux:
## Screenshots

### Before

screencapture-localhost-user-login-2019-11-19-13_07_34


### After

screencapture-localhost-user-login-2019-11-19-13_07_56

app/templates/_base_page.html
```diff
@@ -18,6 +18,14 @@
     {% endblock %}
     <main id="content" role="main">
       {% include "toolkit/flash_messages.html" %}
+      {% block errorSummary %}
+        {% if errors %}
+          {{ govukErrorSummary({
+          "titleText": "There is a problem",
```
katstevens: Interesting that 'There is a problem' tested well with users (according to the [Design System issue](https://github.com/alphagov/govuk-design-system-backlog/issues/46)! It feels a little unnatural to me. I wonder if we'll run into scenarios in other apps where we need to overwrite it with something else...

lfdebrux: We've put the error summary into a block so it can be overridden for whatever reason; however I would like to think that this will be needed only rarely.

In general I think we should be moving our error messages to be based to Design System recommendations wherever possible, to reduce the amount of content we have to manage in our service.

aliuk2012: I think it would be good to get @NoraGDS to review and see if we can update the content. I have read some guidance somewhere on in the design system about trying not to blame the user. For example the existing error messages make me feel like I've done something terrible and it comes across as forceful ("You must..."). Design System do have patterns regarding error messages for specific things, for example, email address error messages (https://design-system.service.gov.uk/patterns/email-addresses/#error-messages)

lfdebrux: @aliuk2012 yes, the plan is to remove the "you must" messages 👍

NoraGDS: Yes, that would be my recommendation too. So if the 4 'You must provide' are removed and replaced with 'Enter', that would comply with the Design System too.

katstevens:
On the reset password form, should there be a red margin on the left hand side? Also the 'You must provide an email address' isn't red...

reset-pw-error-margin

katstevens:
The margin isn't showing up for me on the other templates either (it's there on master...)

lfdebrux:
> On the reset password form, should there be a red margin on the left hand side? Also the 'You must provide an email address' isn't red...

This appears to be a JavaScript problem, adding `validation.js` back in to the application restores the functionality.

I have no idea why this is.

lfdebrux:
> On the reset password form, should there be a red margin on the left hand side? Also the 'You must provide an email address' isn't red...

Restoring `application.scss` fixes this, I'm not sure why this causes issues with this page only

lfdebrux:
Fixed now.

aliuk2012:
> The margin isn't showing up for me on the other templates either (it's there on master...)

Good spot @katstevens, clearly validation.css wasn't purely affecting just the "validation" component.

PR #191, opened 2019-11-18
Update textboxes to use GOV.UK Frontend text input component

aliuk2012:
We should be able to use `govuk-frontend` input component. This
replaces `digitalmarketplace-frontend-toolkit` textbox component.

We still need to decide whether we still want to have bold and slightly
larger labels compared to govuk-frontend default.

## Before

Screenshot 2019-11-18 at 18 41 12

## After

image

Ticket: https://trello.com/c/aSN47QjR

### DIFF
--- a/app/templates/auth/change-password.html
+++ b/app/templates/auth/change-password.html
@@ -29,36 +29,48 @@
       <div class="govuk-grid-column-two-thirds">
         {{ form.hidden_tag() }}
 
-        {% with question = form.old_password.label,
-                name = "old_password",
-                error = errors.get('old_password', {}).get('message'),
-                value = form.old_password.data,
-                type = "password",
-                autocomplete = "off"
-        %}
-          {% include "toolkit/forms/textbox.html" %}
-        {% endwith %}
+        {{ govukInput({
+          "label": {
+            "text": form.old_password.label.text,
+          },
+          "errorMessage": errors.old_password.errorMessage,
+          "id": "input-old_password",
+          "type": "password",
+          "name": "old_password",
+          "value": form.old_password.data,
+          "attributes": {
+            "autocomplete": "off",
+          },
+        }) }}
 
-        {% with question = form.password.label,
-                name = "password",
-                error = errors.get('password', {}).get('message'),
-                hint = form.password.hint,
-                value = form.password.data,
-                type = "password",
-                autocomplete = "off"
-        %}
-          {% include "toolkit/forms/textbox.html" %}
-        {% endwith %}
+        {{ govukInput({
+          "label": {
+            "text": form.password.label.text,
+          },
+          "errorMessage": errors.password.errorMessage,
+          "id": "input-password",
+          "type": "password",
+          "name": "password",
+          "value": form.password.data,
+          "attributes": {
+            "autocomplete": "off",
+          },
+        }) }}
 
-        {% with question = form.confirm_password.label,
-                name = "confirm_password",
-                error = errors.get('confirm_password', {}).get('message'),
-                value = form.confirm_password.data,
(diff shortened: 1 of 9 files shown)


lfdebrux:
I feel like for the log in page at least we should have bold and large labels for the inputs; here is a comparison of the different options

### Normal

Screenshot 2019-11-20 at 10 41 51

### Bold

Screenshot 2019-11-20 at 10 33 17

### Bold and font size 24

Screenshot 2019-11-20 at 10 34 06

lfdebrux:
Thinking about it further, I think we should use bold and large labels for all form components until we have finished migration; otherwise we will have inconsistent styles across our apps (and potentially between different pages in the same app, depending on how we distribute the work of updating forms).

risicle:
I concur.

lfdebrux:
Okay, we've got some overrides, although we're not quite ready to put them in digitalmarketplace-govuk-frontend.

This is what the login page looks like now:

screencapture-localhost-user-login-2019-11-21-15_57_43

lfdebrux:
I've tested this with the functional tests locally and there is one small change required: https://github.com/alphagov/digitalmarketplace-functional-tests/pull/708

lfdebrux:
https://github.com/alphagov/digitalmarketplace-functional-tests/pull/708 has now been merged, so this PR is ready to be merged once it has been approved.

app/templates/auth/request-password-reset.html
```diff
@@ -33,15 +33,23 @@ <h1 class="govuk-heading-xl">Reset password</h1>
...
+            "text": form.email_address.hint,
+          },
+          "errorMessage": errors.get('email_address', {}).get('errorMessage'),
+          "id": "input-email_address",
+          "type": "email",
+          "name": "email_address",
+          "value": form.email_address.data,
+          "attributes": {
+            "autocomplete": "off",
+            "spellcheck": "false",
```
katstevens: Do we only need `spellcheck: false` for email fields? What does this gain us?

lfdebrux: This is based on advice from GOV.UK Design System here https://design-system.service.gov.uk/patterns/email-addresses/#help-users-to-enter-a-valid-email-address:

> setting the spellcheck attribute to false so that browsers do not spellcheck the email address

I don't know how big of a difference it makes, as according to [the MDN article on spellcheck](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/spellcheck) it's not clear whether most mainstream browsers spellcheck input fields. I can't see including it doing any harm however.

If you want I'm happy to take it out.

katstevens: If that's the guidance then I am happy 👍

app/templates/auth/change-password.html
```diff
@@ -29,36 +29,48 @@ <h1 class="govuk-heading-xl">Change your password</h1>
...
-                type = "password",
-                autocomplete = "off"
-        %}
-          {% include "toolkit/forms/textbox.html" %}
-        {% endwith %}
+        {{ govukInput({
+          "label": {
+            "text": form.old_password.label.text,
+          },
+          "errorMessage": errors.get('old_password', {}).get('errorMessage'),
```
katstevens: Can we now simplify these to: 

`"errorMessage": errors.old_password.errorMessage,`

We have that pattern as an example in `dmutils.forms.helpers` - if we still need the old `get('name', {})` then we should change the example there...

lfdebrux: 👍 I'll simplify these

lfdebrux: Done

PR #192, opened 2019-11-21
Fix hardcoded error templates

katstevens:
https://trello.com/c/yH2to2DG/1137-error-templates-not-consistently-being-used

While testing something else, I spotted that a couple of the views were directly rendering the old FE toolkit error templates, with the old heading style:

500-user-fe

Changes:

- `reset_password.py` now simply calls `abort(503)` instead of rendering a specific error template (and passing it an error message). Instead we log the error internally ('email not sent') and show the  standard 'Try again later' message to the user, which makes it consistent with our other 5xx error behaviour and also gives a proper call to action to the user.
- `create_user.py` now uses the `dmutils.errors.render_error_page` helper to display custom error messages for 400 errors. This helper looks for the new digitalmarketplace-govuk-frontend template folder first, and only falls back to the old toolkit if it can't find it. 

Opinions sought: does the heading does look a little *too* big on the new template now? Should it be 2/3rds width as per the [Design System error page patterns](https://design-system.service.gov.uk/patterns/page-not-found-pages/)?  

user-fe-400-token-error


app/main/views/create_user.py
```diff
@@ -30,10 +31,11 @@ def create_user(encoded_token):
...
             extra={'encoded_token': encoded_token}
         )
-        return render_template(
-            "toolkit/errors/400.html",
+        # Replace this with a proper dedicated template.
+        return render_error_page(
+            status_code=400,
             error_message=INVALID_TOKEN_MESSAGE.format(support_email=current_app.config['SUPPORT_EMAIL_ADDRESS']),
-        ), 400
+        )
```
lfdebrux: Nice 👍

PR #199, opened 2019-12-13
Change footer layout

lfdebrux:
Ticket: https://trello.com/c/lXpy2vZH/142-3-change-footer-to-use-two-column-layout-in-digitalmarketplace-frontend-toolkit-and-all-frontend-apps

## Screenshots
after

http://localhost/user/login


aliuk2012:
Any idea why the frontend had its own footer.scss? I seem to recall there being a slight difference. it might have been just been fixing linting issues between the two i.e adding missing semi-colons and code style.

app/assets/scss/application.scss
```diff
@@ -39,7 +40,6 @@ $govuk-compatibility-govukelements: true;
 @import "node_modules/digitalmarketplace-govuk-frontend/govuk-frontend/all";
 
 // App specific styles
-@import "_footer.scss";
 @import "_text.scss";
 @import "_secondary-action-link.scss";
```
aliuk2012: Can we remove this? We aren't using secondary-action-link in this repo

app/assets/javascripts/application.js
```diff
@@ -5,7 +5,6 @@
   directives to concatenate multiple Javascript files into one.
 */
 //= require ../../../node_modules/jquery/dist/jquery.js
-//= require ../../../node_modules/hogan.js/web/builds/3.0.2/hogan-3.0.2.js
 //= require ../../../node_modules/digitalmarketplace-frontend-toolkit/toolkit/javascripts/option-select.js
 //= require ../../../node_modules/digitalmarketplace-frontend-toolkit/toolkit/javascripts/support.js
 //= require ../../../node_modules/digitalmarketplace-frontend-toolkit/toolkit/javascripts/list-entry.js
```
aliuk2012: Could we also remove list-entry.js and list-entry.scss as we do not use it in this repo?

lfdebrux:
> Any idea why the frontend had its own footer.scss? I seem to recall there being a slight difference. it might have been just been fixing linting issues between the two i.e adding missing semi-colons and code style.

There wasn't any difference, I checked with `diff` before removing.

app/assets/scss/application.scss
lfdebrux: I think so? This should be in a separate PR though

app/assets/javascripts/application.js
lfdebrux: As above, I think this should be in a separate PR.

PR #205, opened 2020-01-07
Try Dependabot

lfdebrux:
I would like to try using [Dependabot](https://dependabot.com) with a frontend repo, as we have been having issues with PyUp recently. This PR is an experiment, I expect that it will either be superseded or reverted at a later date.

This PR adds a [Dependabot config file](https://dependabot.com/docs/config-file/) that I think should cause Dependabot to update requirements in `requirements.txt` weekly. Unfortunately the configuration for Dependabot doesn't seem to be as flexible as PyUp's. I can't see a way to specify filenames using the Dependabot config file format; there might be a way to do so in the GUI, I'm not sure.

On the plus side for Dependabot, we could use it to manage both Python and JavaScript dependencies. Also, Dependabot claims that its [Python features](https://dependabot.com/python/) include being able to manage Git sources (not sure how).

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -1,0 +1,14 @@
+# Dependabot config
+# https://dependabot.com/docs/config-file/
+# Checks and updates dependencies in
+# `requirements.in` and `requirements.txt`
+
+version: 1
+update_configs:
+  - package_manager: "python"
+    directory: "/"
+    update_schedule: "weekly"
+    allowed_updates:
+      - match:
+          dependency_type: "production"
+          update_type: "all"


katstevens:
What do we expect the resulting workflow to be when a PR is raised? I'm assuming a dev would have to manually add to `requirements-app` and refreeze?

lfdebrux:
> What do we expect the resulting workflow to be when a PR is raised? I'm assuming a dev would have to manually add to `requirements-app` and refreeze?

I think so, yeah. I'm not sure though, the dependabot docs are pretty sparse. One thing that's interesting is that if we were using `pip-tools` then Dependabot supposedly would pick that up and work with that; I wonder if changing the names/layout of our requirement files to look like they were generated with `pip-compile` would make Dependabot do the freezing for us?

lfdebrux:
> I wonder if changing the names/layout of our requirement files to look like they were generated with `pip-compile` (i.e. name it `requirements.in` like this [blog post](https://dependabot.com/blog/dependabot-now-supports-pip-compile/) mentions) would make Dependabot do the freezing for us?

Looking at the [dependabot-core source code](https://github.com/dependabot/dependabot-core/blob/master/python/lib/dependabot/python/file_updater/pip_compile_file_updater.rb) it seems that Dependabot uses `pip-compile` if it detects a `requirements.in` file, so the above suggestion is probably a bad one.

katstevens:
How annoying would it be for us to swap our filenames? So that `requirements.txt` becomes `requirements-app.txt` and vice-versa? That would give us the equivalent of the PyUp PR, wouldn't it?

lfdebrux:
> How annoying would it be for us to swap our filenames? So that `requirements.txt` becomes `requirements-app.txt` and vice-versa? That would give us the equivalent of the PyUp PR, wouldn't it?

That might work 🤔 Although I'm reading the source code for Dependabot trying to figure out what it does exactly and I'm starting to worry that actually it looks at all `*.txt` files...

lfdebrux:
Okay, so my theory on reading the code is that the config in this PR will cause Dependabot to update both `requirements-app.txt` and `requirements.txt`. I've update the comment in the config to reflect this. We could get around this if we changed the file extension of `requirements-app.txt` (i.e. to something like `requirements-app.lst`, or just `requirements-app`). We as developers would  then have to freeze requirements manually for Dependabot PRs, like we do for PyUp PRs.

Alternatively, it might be okay to let Dependabot update both files, depending on whether/how Dependabot looks at version conflicts. I will keep researching.

lfdebrux:
After a lot of reading, it looks like Dependabot doesn't do conflict resolution of its own (which is the reason why we do `make freeze`.

lfdebrux:
I think this would work better if we were using `pip-tools`; I've put up a PR that implements that if anyone is interested #207

lfdebrux:
WIPing until we make a decision on whether to go with this

PR #218, opened 2020-02-03
Add script to test whether `pip-compile` needs to be run

lfdebrux:
Ticket: https://trello.com/c/iLQca47V/251-replace-custom-script-to-freeze-requirements-with-pip-tools

We want to be reminded to run `pip-compile` when changing `requirements.in` or `requirements-dev.in`.

This works by a three step process:

1. Has the "infile" (i.e. `requirements.in` or `requirements-dev.in`) changed?
2. Has the requirements file changed?
3. If the infile has changed but the requirements file hasn't, run `pip-compile` and see if the requirement file should change

Running `pip-compile` is slow, which is why we first test the infile. We test for changes by comparing with the git master branch.

This is a bit of a poor test and probably will mess up in future, but it might be better than nothing.


lfdebrux:
Travis failure shows how this script works :D

image

phantas:
how long does this take to run? have you considered adding a pre-commit so the freeze is run whenever requirements.in or requirements.dev.in is changed?

lfdebrux:
> how long does this take to run?

`make freeze-requirements` runs in just under a minute. This script takes less than half a second if it it doesn't need to freeze requirements.

> have you considered adding a pre-commit so the freeze is run whenever requirements.in or requirements.dev.in is changed?

Yeah I thought about it, however one use case for this test is preventing devs from merging PRs from PyUp without first running `pip compile`; I don't think a pre-commit hook would catch that?

scripts/test_requirements.sh
```diff
@@ -0,0 +1,71 @@
...
+    status=1
+  fi
+done
+
+if [ $status -ne 0 ]
+then
+  # we want to allow changes to infiles that don't affect the requirements file
+  # the easiest way to do this is to run `pip-compile` and test the output
+  >&2 echo 'Running `make freeze-requirements`...'
+  make freeze-requirements > /dev/null 2>&1
```
phantas: this is a bit of a loop: Makefile -> Bash -> Makefile, how about having on the makefile something like

```
freeze-requirements: virtualenv requirements-dev requirements.in
	./scripts/test_requirements.sh || ${VIRTUALENV_ROOT}/bin/pip-compile
	./scripts/test_requirements.sh dev || ${VIRTUALENV_ROOT}/bin/pip-compile requirements-dev.in
```
drop the test-requirements, and just update the Makefile dependencies accordingly

(you can also have an if block instead, if you prefer to have the test requirements run for both instead of per file)

phantas:
> > how long does this take to run?
> 
> `make freeze-requirements` runs in just under a minute. This script takes less than half a second if it it doesn't need to freeze requirements.
> 
> > have you considered adding a pre-commit so the freeze is run whenever requirements.in or requirements.dev.in is changed?
> 
> Yeah I thought about it, however one use case for this test is preventing devs from merging PRs from PyUp without first running `pip compile`; I don't think a pre-commit hook would catch that?

No, I don't think it would, but you can add it to Travis too. It just helps if two devs are working on the same branch and the app dependencies to be consistent, particularly for long living branches

scripts/test_requirements.sh
```diff
@@ -0,0 +1,71 @@
...
+# Requirements files to check
+REQUIREMENTS_FILES=${@:-requirements.txt requirements-dev.txt}
+
+# Git reference to compare against
+BASE_COMMIT=${BASE_COMMIT:-master}
+HEAD_COMMIT=${HEAD_COMMIT:-HEAD}
+
+# Return 1 if file(s) has changed relative to BASE_COMMIT, otherwise 0
+test_unchanged() {
+  git diff --quiet $BASE_COMMIT $HEAD_COMMIT -- $@
```
phantas:
```suggestion
  git diff --quiet $BASE_COMMIT $HEAD_COMMIT -- "$@"
```

Tiny suggestion: guard against spaces in the file name.

phantas:
```suggestion
  git diff --quiet $BASE_COMMIT $HEAD_COMMIT -- "$@"
```

Tiny suggestion: guard against spaces in filename. But you'd need to change the rest of this script too for any mention of the file so happy if you disregard

scripts/test_requirements.sh
```diff
@@ -0,0 +1,71 @@
...
+}
+
+# Return filename for requirements file
+requirements_infile() {
+  echo ${1%%.txt}.in
```
phantas: ah gawds how I hate the workaround for treating bash procedures as functions 😄 no suggestions, just whining.

scripts/test_requirements.sh
lfdebrux: In this case I want the spaces to be expanded so that multiple files can be supplied to the command, for instance `test_changed file1 file2` should check whether file1 or file2 had been changed.

I'll add a usage line to make this clearer

scripts/test_requirements.sh
phantas: Oh!, I hadn't quite realised what you were doing here... although the same principle still applies.
Having a changed file that was not committed reinforces a bit my case for having a pre-commit hook too, I for one don't quite trust any checks that rely on humans. I need to do some thinking about how to pair this with dependabots though...

scripts/test_requirements.sh
phantas: That makes sense. Sorry, duplicate comment because I pressed the wrong button

lfdebrux:
> This is great! I want to do some thinking about integration with dependabots.

Dependabot has pip-tools support built in, so this hack shouldn't be needed for it 😁

phantas:
I am just thinking: won't dependabots update the requirements.txt file instead of the requirements.in . Travis will say that requirements.txt needs changing, where in fact it's requirements.in, which causes a degree of confusion at best.

lfdebrux:
> I am just thinking: won't dependabots update the requirements.txt file instead of the requirements.in . Travis will say that requirements.txt needs changing, where in fact it's requirements.in, which causes a degree of confusion at best.

If Dependabot detects a `*.in` file it runs `pip-compile` under the hood (it's not documented, but I found this by looking at the [source code](https://github.com/dependabot/dependabot-core/blob/master/python/lib/dependabot/python/file_updater/pip_compile_file_updater.rb)), so in that case this test should pass.

benvand:
Having this as a blocking precommit hook will break some workflows. It is sometimes useful to be able to put up a PR before it is possible to lock all the requirements.


For example:
1. I put up a PR against git+https://github.com/alphagov/digitalmarketplace-apiclient.git with a minor version bump adding a new client method.
2. I make the changes to this repo which require the new digitalmarketplace-apiclient version and client method
3. I commit the changes to my branch and open a PR on this repo, using the new client method and with the new version number in this repos `requirements.in`

Equally if it's non blocking we need to surface it at some point. Logical point is the travis tests.

benvand:
Honestly, would rather see this as a python module in https://github.com/alphagov/digitalmarketplace-utils/blob/master/dmutils/repoutils
😬 

We'll need to bring this in to all the apps right?
And we already bring utils in to all the apps.

Just a straight conversion of this from bash to python? And be able to run it with
`python -m dmutils.repoutils.something-something`

lfdebrux:
> Honestly, would rather see this as a python module in https://github.com/alphagov/digitalmarketplace-utils/blob/master/dmutils/repoutils
> 😬
> 
> We'll need to bring this in to all the apps right?
> And we already bring utils in to all the apps.
> 
> Just a straight conversion of this from bash to python? And be able to run it with
> `python -m dmutils.repoutils.something-something`

I'll give it a try...

atugushev:
Just passing by :) I use pre-commit with this `.pre-commit-config.yaml`:

```yml
  - repo: local
    hooks:
      - id: pip-compile
        name: pip-compile
        entry: pip-compile
        language: system
        files: ^requirements.(in|txt)$
        pass_filenames: false
```

Works for me!

lfdebrux:
Thanks for the thoughts and suggestions everyone.

I'm going to close this PR while I look at options.

PR #255, opened 2020-04-30
Use mainline pip-tools instead of fork

lfdebrux:


katstevens:
Weird that this one is failing on Travis - no `pip-check` available?

lfdebrux:
> Weird that this one is failing on Travis - no `pip-check` available?

Oh woops forgot to remove something from the Makefile. Thanks for flagging up.

PR #258, opened 2020-05-07
use project local version of gulp

gidsg:
when running `dm-runner` `make setup` on a clean machine I got the following error:
```
16:51:51            user-frontend | npm run --silent frontend-build:development
16:51:52            user-frontend | sh: gulp: command not found
```
this is because the `package.json` assumed a global installation of gulp-cli. 
This PR changes it to use the project bundled version of gulp which should resolve this issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,9 +21,9 @@
     "gulp-uglify": "^3.0.2"
   },
   "scripts": {
-    "frontend-build:development": "gulp build:development",
-    "frontend-build:production": "gulp build:production",
-    "frontend-build:watch": "gulp watch",
+    "frontend-build:development": "./node_modules/gulp/bin/gulp.js build:development",
+    "frontend-build:production": "./node_modules/gulp/bin/gulp.js build:production",
+    "frontend-build:watch": "./node_modules/gulp/bin/gulp.js watch",
     "test": "standard \"gulpfile.js\""
   },
   "devDependencies": {


katstevens:
Was this an issue just for the User Frontend, or were other apps affected?

gidsg:
@katstevens 
> Was this an issue just for the User Frontend, or were other apps affected?
only for this app.

katstevens:
> @katstevens
> 
> > Was this an issue just for the User Frontend, or were other apps affected?
> > only for this app.

Hmm I'd like to know why User FE is special, before we make this app diverge from the others?

gidsg:
@katstevens 
> I'd like to know why User FE is special, before we make this app diverge from the others?

It appears we do already use this local style in other apps e.g. https://github.com/alphagov/digitalmarketplace-frontend-toolkit/blob/323b3d4b704e33782df6cdae7b105a2a9237a594/package.json#L16-L18
In other places we do not use it e.g.
https://github.com/alphagov/digitalmarketplace-design-patterns/blob/ca4094111cb4a9a3b70d47f4131e18f72747744f/package.json#L12
however, I guess these cases are not invoked by `make setup`

I would say this style is best practice i.e. it is used by notify in:
https://github.com/alphagov/notify-frontend/blob/509695a0e462bef711bc73de3978aa0683f5603a/package.json#L21

lfdebrux:
According to gulp it's not best practice 🤔 

https://gulpjs.com/docs/en/getting-started/quick-start#install-the-gulp-command-line-utility

gidsg:
@lfdebrux OK, I noticed notify also don't do this in their current admin app https://github.com/alphagov/notifications-admin/blob/master/package.json#L11
so would the solution be for `dm-runner` to install `gulp` globally? Or should we document that as a manual step? Looks like a similar case for `standard` which we invoke globally.

lfdebrux:
Hmm, it would be nice if dmrunner dealt with this for us, but it also feels like something that should also be documented 😆

gidsg:
OK, I will close this as-per discussion and look to add something to the manual initially instead.

PR #261, opened 2020-05-21
add link to error summary box

gidsg:
Add [error summary component](https://design-system.service.gov.uk/components/error-summary/)  link.
This is part of: https://trello.com/c/btdB5xBK/98-missing-contextual-error-messaging-on-cookie-settings-page
Screenshot 2020-05-21 at 15 06 28

### DIFF
--- a/app/templates/cookies/cookie_settings.html
+++ b/app/templates/cookies/cookie_settings.html
@@ -25,7 +25,11 @@
     "type": "error",
     "classes": "dm-cookie-settings__error",
     "titleText": "There was a problem saving your settings",
-    "text": "Please select 'Yes' or 'No'.",
+    "html": '<ul class="govuk-list govuk-error-summary__list">
+      <li>
+        <a href="#dm-cookie-settings">Please select \'Yes\' or \'No\'.</a>
+      </li>
+    </ul>',
     "attributes": [("id", "dm-cookie-settings-error")]
   }) }}
 
--- a/app/templates/cookies/cookie_settings.html
+++ b/app/templates/cookies/cookie_settings.html
@@ -29,7 +29,7 @@
     },
     "errorList": [
       {
-        "text": "Please select 'Yes' or 'No'",
+        "text": "Select yes to accept analytics cookies",
         "href": "#dm-cookie-settings"
       }
     ]
--- a/app/templates/cookies/cookie_settings.html
+++ b/app/templates/cookies/cookie_settings.html
@@ -21,16 +21,18 @@
 
 {% block mainContent %}
 
-  {{ dmAlert({
-    "type": "error",
-    "classes": "dm-cookie-settings__error",
+  {{ govukErrorSummary({
     "titleText": "There was a problem saving your settings",
-    "html": '<ul class="govuk-list govuk-error-summary__list">
-      <li>
-        <a href="#dm-cookie-settings">Please select \'Yes\' or \'No\'.</a>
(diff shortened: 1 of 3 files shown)


app/templates/cookies/cookie_settings.html
```diff
@@ -25,7 +25,11 @@
...
     "classes": "dm-cookie-settings__error",
     "titleText": "There was a problem saving your settings",
-    "text": "Please select 'Yes' or 'No'.",
+    "html": '<ul class="govuk-list govuk-error-summary__list">
+      <li>
+        <a href="#dm-cookie-settings">Please select \'Yes\' or \'No\'.</a>
+      </li>
+    </ul>',
     "attributes": [("id", "dm-cookie-settings-error")]
   }) }}
```
katstevens, line 34: We might actually be able to replace this `dmAlert` with the [govukErrorSummary component](https://design-system.service.gov.uk/components/error-summary/): 

```
{% from "components/error-summary/macro.njk" import govukErrorSummary %}

{{ govukErrorSummary({
      "titleText": "There was a problem saving your settings",
      "errorList": [{
        "text": "Please select \'Yes\' or \'No\'.",
        "href": "#dm-cookie-settings"
      }],
      "classes": "dm-cookie-settings__error" 
    }) }}
```

app/templates/cookies/cookie_settings.html
```diff
@@ -21,12 +21,18 @@
...
 {% block mainContent %}
 
-  {{ dmAlert({
-    "type": "error",
-    "classes": "dm-cookie-settings__error",
+  {{ govukErrorSummary({
     "titleText": "There was a problem saving your settings",
-    "text": "Please select 'Yes' or 'No'.",
-    "attributes": [("id", "dm-cookie-settings-error")]
+    "classes": "dm-cookie-settings__error",
```
domoscargin, line 26: We should either remove this class or get rid of the @include in its scss (which is using toolkit styles which don't affect the Error Summary component): https://github.com/alphagov/digitalmarketplace-user-frontend/blob/65b0b340ec9d5554c9ed22ae988b5c56bece97af/app/assets/scss/overrides/_cookie-settings.scss#L24-L27

To remove it, looks like we'd have to call [this function](https://github.com/alphagov/digitalmarketplace-govuk-frontend/blob/eb4c7fc9e9d0732421cce8307d181aab6a740388/src/digitalmarketplace/components/cookie-settings/cookie-settings.js#L105-L111 ) on load, so maybe better to keep the display: none override.

app/templates/cookies/cookie_settings.html
```diff
@@ -21,12 +21,18 @@
...
+     "attributes": {
+      "id": "dm-cookie-settings-error"
+    },
+    "errorList": [
+      {
+        "text": "Please select 'Yes' or 'No'",
```
domoscargin, line 32: According to the [Design System](https://design-system.service.gov.uk/components/radios/#error-messages), this should be something like "Select yes to accept analytics cookies"
```

NoraGDS: If following Design System, it should be: Select yes if you want to accept analytics cookies. This seems a bit long and could be even simpler, so agree with @domoscargin - Select yes to accept analytics cookies

lfdebrux:
Looks good to me.

I like the way this can be deployed independently of alphagov/digitalmarketplace-govuk-frontend#124.

PR #265, opened 2020-06-09
Replace flash messages with Digital Marketplace GOV.UK Frontend dmAlert component

lfdebrux:
Ticket: https://trello.com/c/sxhsW20p/88-replace-flash-messages-with-alert-component-in-briefs-frontend

Flash messages from the user frontend often end up showing up in the briefs frontend, so we want to make the change here first.


app/templates/_base_page.html
```diff
@@ -49,7 +49,20 @@
 
 
 {% block content %}
-  {% include "toolkit/flash_messages.html" %}
+  {% block flashMessages %}
+    {% with
+       messages = get_flashed_messages(with_categories=True),
+       titles = {"error": "There is a problem", "success": "Success"}
```
katstevens, line 55: Is "Success" now going to be displayed above the usual success `message` as a title?

tests/main/views/test_reset_password.py
```diff
@@ -435,7 +435,7 @@ def test_token_created_before_last_updated_password_cannot_be_used(self):
 
         assert res.status_code == 200
         document = html.fromstring(res.get_data(as_text=True))
-        error_selector = cssselect.CSSSelector('div.banner-destructive-without-action')
+        error_selector = cssselect.CSSSelector('div.dm-alert.dm-alert--error')
```
katstevens, line 438: Can these assertion lines be handled with `self.assert_flashes(..., 'error')` as per the other tests?

app/templates/_base_page.html
lfdebrux: Yes.

It should look something like this:

Success. You have successfully changed your password.

katstevens: Interesting! Is this a Design System pattern/recommendation, or our own decision?

tests/main/views/test_reset_password.py
lfdebrux: Not without rewriting the test quite a bit.

`assert_flashes()` checks for the presence of the flash message in the session data. The flash message is popped from the session data when it is shown, which I think is why the other tests don't follow the redirect, so the page showing the flash message is never reached.

Since this test uses `follow_redirects=True` when POSTing the data, the flash message is shown, and `assert_flashes()` raises an error `nothing flashed`. I'm not sure why this test is different to the others, but looking at the test name `test_token_created_before_last_updated_password_cannot_be_used` I think the flash message test might be entirely incidental to the actual thing being tested...

katstevens: Ahh it's one of those. Fair enough.

app/templates/_base_page.html
lfdebrux: Now that I think about it, I can't remember 😅

I think it might have been something I did ages ago when doing a spike into alerts without thinking too hard about it.

Maybe we should take it out then?

PR #272, opened 2020-07-13
Review error message content

katstevens:
https://trello.com/c/ITSfGrTU/126-1-review-error-message-content-in-user-frontend

Going through the User FE's error messages, of which there are many combinations of subtly different errors.

- Email field is empty
- Email field does not contain a valid email
- Password field is empty (when logging in)
- Password field is empty (when changing an existing password)
- Password is the wrong length (when changing an existing password)
- Password is incorrect (when logging in - TODO: get input from Content Designer)
- Password is incorrect (when changing an existing password)
- Password is too easy to guess*
- Password change succeeded
- Password change failed

*While updating this error type, I realised we were still using the outdated whitelist/blacklist terminology in this repo. So I added a commit to move to the more inclusive safelist/blocklist.


app/main/forms/auth_forms.py
```diff
@@ -28,6 +28,7 @@
 PASSWORD_MISMATCH_ERROR_MESSAGE = "The passwords you entered do not match"
 NEW_PASSWORD_EMPTY_ERROR_MESSAGE = "Enter a new password"
 NEW_PASSWORD_CONFIRM_EMPTY_ERROR_MESSAGE = "Confirm your new password"
+PASSWORD_CHANGE_AUTH_ERROR_MESSAGE = "Make sure you’ve entered the right password."
```
NoraGDS, line 31: Would this work?: Enter your old password

katstevens: Sounds good to me! Will edit, thanks Nora 👍

app/main/views/reset_password.py
```diff
@@ -28,8 +28,8 @@
     """
 )
 
-PASSWORD_UPDATED_MESSAGE = "You have successfully changed your password."
-PASSWORD_NOT_UPDATED_MESSAGE = "Could not update password due to an error."
+PASSWORD_UPDATED_MESSAGE = "Your password has been successfully changed."
```
domoscargin, line 31:
```suggestion
PASSWORD_UPDATED_MESSAGE = "Your password has been changed successfully."
```

I can never remember what's right (_something something split infinitive something_?), but this sounds more correct to me.

Googling the suggested phrase brings up 18 million+ results vs the existing phrase with 250K.

katstevens: Interesting! It felt more natural to me as well.

PR #204, opened 2019-12-24
🚀 Replacing govuk_template ☠️

aliuk2012:
## Blocked by
- Phase banner & User research banner have been replaced with govuk-frontend phase banner component in all the other applications.

## What has changed?
- Removes dependency on GOV.UK Template
- Uses GOV.UK Frontend template
- Removes temporary hacks needed to ensure a smooth transition

## Other improvements
- Removes all JavaScript not needed by digitalmarketplace-user-frontend
- Removes dependency on **jQuery** 😍

## Performance impact:

|                           | Current | New   | % difference  |
|---------------------------|---------|-------|---------------|
| Requests                  | 13      | 7     | 46% reduction | 
| Transferred (Gzip)        | 467KB   | 172KB | 63% reduction |
| Transferred (Uncompressed)| 729KB   | 295KB | 60% reduction |

## Before screenshot
image

## After screenshot
image

## Lighthouse Reports (Snapshot):
🤮[Preview](https://googlechrome.github.io/lighthouse/viewer/?gist=c871de1340b97c6b60841436a26e843f)
🤩[Localhost](https://googlechrome.github.io/lighthouse/viewer/?gist=e23cae940c3dfe2f7bb4a81b0cda765f)

Ticket: https://trello.com/c/lGqdZbhS

PR #275, opened 2020-08-03
Migrate to govuk-frontend V3

domoscargin:
https://trello.com/c/5FhjhxMd/139-3-migrate-user-frontend-to-govuk-frontend-v3


app/main/views/notifications.py
```diff
@@ -48,23 +47,9 @@ def user_research_consent():
         user = data_api_client.get_user(current_user.id)
         form = UserResearchOptInForm(user_research_opt_in=user['users']['userResearchOptedIn'])
 
-    # Set the seen_user_research_message cookie if it does not exist.
```
gidsg, line 51: not sure of the history here. @katstevens can you confirm this is safe to delete?

package.json
```diff
@@ -8,10 +8,7 @@
   },
   "dependencies": {
     "del": "^5.1.0",
-    "digitalmarketplace-frontend-toolkit": "https://github.com/alphagov/digitalmarketplace-frontend-toolkit.git#v36.1.0",
-    "digitalmarketplace-govuk-frontend": "^0.9.1",
-    "govuk-elements-sass": "3.0.3",
-    "govuk_frontend_toolkit": "5.0.3",
+    "digitalmarketplace-govuk-frontend": "github:alphagov/digitalmarketplace-govuk-frontend#26080ba",
```
gidsg, line 11:
```suggestion
    "digitalmarketplace-govuk-frontend": "github:alphagov/digitalmarketplace-govuk-frontend#v1.0.0",
```
Can we point to tag rather than commit hash?

.gitignore
```diff
@@ -5,8 +5,6 @@ venv*
 node_modules
 npm-debug.log
 app/static/*
-govuk_template
-digital_marketplace_frontend_toolkit
```
gidsg, line 9: ❤️

app/assets/scss/application.scss
```diff
@@ -1,27 +1,8 @@
...
-@import "toolkit/shared_placeholders/_mixins.scss";
-
-
-// Digital Marketplace Front-end toolkit styles (only import the files you need)
-@import "toolkit/_notification-banners.scss";
-@import "toolkit/_proposition-header.scss";
-@import "toolkit/forms/_hint.scss";
-@import "toolkit/forms/_questions.scss";
-@import "toolkit/forms/_textboxes.scss";
-@import "toolkit/forms/_selection-buttons.scss";
```
lfdebrux, line 15: 🔥

tests/helpers.py
```diff
@@ -254,9 +254,7 @@ def assert_flashes(self, expected_message, expected_category='message'):
             assert expected_category == category
 
     def assert_breadcrumbs(self, response, extra_breadcrumbs=None):
-        breadcrumbs = html.fromstring(response.get_data(as_text=True)).xpath(
-            '//*[@class="govuk-breadcrumbs"]/ol/li'
-        )
+        breadcrumbs = html.fromstring(response.get_data(as_text=True)).cssselect('.govuk-breadcrumbs ol li')
```
lfdebrux, line 257: Much better 😁

gulpfile.js
```diff
@@ -135,42 +126,6 @@ function copyFactory (resourceName, sourceFolder, targetFolder) {
...
-
-gulp.task(
-  'copy:dm_toolkit_assets:templates',
-  copyFactory(
-    'templates from the Digital Marketplace frontend toolkit',
-    dmToolkitRoot + '/templates',
-    'app/templates/toolkit'
-  )
-)
-
```
lfdebrux, line 173: Hallelujah! 👼

app/main/views/notifications.py
lfdebrux: Oh yeah we don't need this we got rid of this in https://github.com/alphagov/digitalmarketplace-user-frontend/pull/204

lfdebrux:
This looks awesome. New focus styles are there, the secondary action link highlights look much better too, and I love the way the page *doesn't* move around when loading!

package.json
domoscargin: I'm happy to - we should update dm-govuk-frontend's `release-to-branch` terminal output to reflect that though : https://github.com/alphagov/digitalmarketplace-govuk-frontend/blob/1ffd3ce34e52a7ad6e49fb128b362b394ea11252/bin/release-to-branch.sh#L44

app/main/views/notifications.py
katstevens: Ditch it! 🔥

app/assets/scss/application.scss
katstevens: I love this commit so much.

app/assets/javascripts/_analytics.js
```diff
@@ -1,3 +0,0 @@
-//= include ../../../node_modules/digitalmarketplace-frontend-toolkit/toolkit/javascripts/analytics/_init.js
-
-//= include analytics/_events.js
```
katstevens, line 3: Gosh I thought we'd already got rid of these! Good stuff

package.json
gidsg: I think release to branch is more for testing but once released we should use a tagged release.

lfdebrux:
### Visual changes

Screenshot 2020-08-03 at 17 44 36

This is an example of the output of the visual regression tests against this commit. Most of the changes are related to colour, although there are a few shifts due to size changes as well.

package.json
lfdebrux: @gidsg I don't think we can do a release, because it's a long-running branch; we need to have changes applied to digitalmarketplace-govuk-frontend to support govuk-frontend v3, and I don't think we want those changes in the default branch.

PR #278, opened 2020-09-04
Install digitalmarketplace-govuk-frontend and govuk-frontend separately

domoscargin:
https://trello.com/c/fhL8Zl8T/175-3-unbundle-govuk-frontend-from-digitalmarketplace-govuk-frontend

digitalmarketplace-govuk-frontend has been updated so that it is no longer bundled with govuk-frontend.

This PR addresses those changes.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -11,7 +11,7 @@ const path = require('path')
 let environment
 const repoRoot = path.join(__dirname)
 const npmRoot = path.join(repoRoot, 'node_modules')
-const govukFrontendRoot = path.join(npmRoot, 'digitalmarketplace-govuk-frontend', 'govuk')
+const govukFrontendRoot = path.join(npmRoot, 'govuk-frontend')
 const assetsFolder = path.join(repoRoot, 'app', 'assets')
 const staticFolder = path.join(repoRoot, 'app', 'static')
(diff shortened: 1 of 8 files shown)


lfdebrux:
### My review checklist

I've:

- [ ] looked at the code
- [ ] tested locally
- [ ] run functional tests locally

gulpfile.js
```diff
@@ -144,6 +144,18 @@ gulp.task(
...
+  function () {
+    return gulp
+      .src(path.join(govukFrontendRoot, '**', '*.scss'), { base: govukFrontendRoot })
+      .pipe(gulp.dest(path.join('app', 'assets', 'scss')))
+      .on('end', function () {
+        console.log('📂  Copied stylesheets from GOV.UK Frontend')
+      })
+  }
+)
+
```
lfdebrux, line 158: Instead of doing this we could add `node_modules/govuk-frontend` to the [include paths in the gulpfile]( https://github.com/alphagov/digitalmarketplace-user-frontend/blob/master/gulpfile.js#L34).

Which do you think is better?

domoscargin: include paths!

lfdebrux (approved):

PR #298, opened 2021-01-25
Add github action

gidsg:


.github/workflows/test.yml
```diff
@@ -0,0 +1,47 @@
...
+      shell: bash -l {0}
+      run: nvm install
+
+    - name: Install Node dependencies
+      run:  make npm-install
+      if: steps.node-cache.outputs.cache-hit != 'true'
+
+    - name: Install python dependencies
+      run: make requirements-dev
+      if: steps.python-cache.outputs.cache-hit != 'true'
```
lfdebrux, line 44: This is cool, however I think `make test` will run those steps anyway?

gidsg: Yes, it does seem to but because the steps are not decoupled, there's no way to avoid that. At least this removes one repetition.

lfdebrux: 👍

lfdebrux (approved):

PR #418, opened 2022-01-04
Add the notification banner

tim-s-ccs:
- Add the notification banner
- Add helper method to determine if DOS6 is live
- Add tests for the banner visibility

This is from the ticket [OST-134](https://crowncommercialservice.atlassian.net/browse/OST-134)

This is a screenshot of what the login page would look like:
localhost_user_login_show_dos6_live=true

I've set it up so the the banner will be shown on and after the DOS6 'Go Live' date (2022-01-14).
In order to see the banner before the 'Go Live' date, a user can add the param `show_dos6_live=true` to the URL.
So for preview, the url to view the banner now would be: https://www.preview.marketplace.team/user/login?show_dos6_live=true

### DIFF
--- a/app/main/helpers/login_helpers.py
+++ b/app/main/helpers/login_helpers.py
@@ -1,9 +1,7 @@
 from urllib.parse import urlparse, urljoin
 
 from flask_login import current_user
-from flask import request, current_app, redirect, url_for
-
-from datetime import date
+from flask import request, redirect, url_for
 
 
 def is_safe_url(next_url):
@@ -47,8 +45,3 @@ def get_user_dashboard_url(current_user_):
         return url_for('external.admin_dashboard')
 
     return None
-
-
-def dos6_live(params):
-    return date.today().strftime('%Y-%m-%d') >= current_app.config['DOS6_GO_LIVE_DATE'] or\
-        params.get('show_dos6_live') == 'true'
--- a/app/main/helpers/login_helpers.py
+++ b/app/main/helpers/login_helpers.py
@@ -3,6 +3,11 @@ from urllib.parse import urlparse, urljoin
 from flask_login import current_user
 from flask import request, redirect, url_for
 
+from datetime import datetime
+
+# Go Live Dates
+DOS6_GO_LIVE_DATE = '2022-01-14'
+
 
 def is_safe_url(next_url):
     """
@@ -45,3 +50,7 @@ def get_user_dashboard_url(current_user_):
         return url_for('external.admin_dashboard')
 
     return None
+
+
+def dos6_live(params):
+    return datetime.utcnow().strftime('%Y-%m-%d') >= DOS6_GO_LIVE_DATE or params.get('show_dos6_live') == 'true'
--- a/app/main/helpers/login_helpers.py
+++ b/app/main/helpers/login_helpers.py
@@ -1,12 +1,9 @@
 from urllib.parse import urlparse, urljoin
 
 from flask_login import current_user
-from flask import request, redirect, url_for
+from flask import request, current_app, redirect, url_for
 
-from datetime import datetime
-
-# Go Live Dates
-DOS6_GO_LIVE_DATE = '2022-01-14'
+from datetime import date
 
 
 def is_safe_url(next_url):
@@ -53,4 +50,5 @@ def get_user_dashboard_url(current_user_):
 
 
 def dos6_live(params):
-    return datetime.utcnow().strftime('%Y-%m-%d') >= DOS6_GO_LIVE_DATE or params.get('show_dos6_live') == 'true'
+    return date.today().strftime('%Y-%m-%d') >= current_app.config['DOS6_GO_LIVE_DATE'] or\
+        params.get('show_dos6_live') == 'true'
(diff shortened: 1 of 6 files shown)


app/main/helpers/login_helpers.py
```diff
@@ -45,3 +50,7 @@ def get_user_dashboard_url(current_user_):
         return url_for('external.admin_dashboard')
 
     return None
+
+
+def dos6_live(params):
+    return datetime.utcnow().strftime('%Y-%m-%d') >= DOS6_GO_LIVE_DATE or params.get('show_dos6_live') == 'true'
```
bjgill, line 56: Do we need to convert `datetime.utcnow()` to a string? Could we compare it as a datetime?

app/templates/auth/login.html
```diff
@@ -19,6 +19,24 @@
...
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-two-thirds">
+        {% set html %}
+          <p class="govuk-notification-banner__heading">
+            Supplier information from DOS6 is not available from this log in screen
+            <a class="govuk-notification-banner__link" href="https://www.applytosupply.digitalmarketplace.service.gov.uk">Start or continue your application for the DOS6 framework</a>
+          </p>
+        {% endset %}
+
+        {{ govukNotificationBanner({
```
bjgill, line 32: ❤️  a new design system component

app/templates/auth/login.html
```diff
@@ -19,6 +19,24 @@
 {% block mainContent %}
   <h1 class="govuk-heading-xl">Log in to the Digital Marketplace</h1>
 
+  {% if dos6_live %}
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-two-thirds">
+        {% set html %}
+          <p class="govuk-notification-banner__heading">
+            Supplier information from DOS6 is not available from this log in screen
```
bjgill, line 27:
```suggestion
            Supplier information from DOS6 is not available from this log in screen.
```

app/main/helpers/login_helpers.py
```diff
@@ -3,6 +3,11 @@
 from flask_login import current_user
 from flask import request, redirect, url_for
 
+from datetime import datetime
+
+# Go Live Dates
+DOS6_GO_LIVE_DATE = '2022-01-14'
```
bjgill, line 9: As this is going to be a ~1 off, I think we're fine to hard-code it.

However, is this the right place for it? Should it live in config?

app/templates/auth/login.html
```diff
@@ -19,6 +19,24 @@
...
+            <a class="govuk-notification-banner__link" href="https://www.applytosupply.digitalmarketplace.service.gov.uk">Start or continue your application for the DOS6 framework</a>
```
bjgill, line 28:
```suggestion
            <a class="govuk-notification-banner__link" href="https://www.applytosupply.digitalmarketplace.service.gov.uk">Start or continue your application for the DOS6 framework.</a>
```

bjgill (commented):
A few small questions, but I like the implementation.

I've got a few spec-level questions, but they're not directly relevant to this PR, so I'll raise them in JIRA

app/main/helpers/login_helpers.py
tim-s-ccs: We could but my thinking was that it might be quicker to compare two strings rather than two date objects? I'm happy to do just do date comparison though as it is probably more appropriate.

app/main/helpers/login_helpers.py
tim-s-ccs: That does seem more sensible the more I think about it, I shall move it in

app/main/helpers/login_helpers.py
bjgill: BTW - I forgot to mention it, but I really like the idea of having a query parameter to allow us to test this.

tim-s-ccs: I'm not sure I really needed to worry about performance as its a simple operation, oh well. I've changed it to date as that's what it should have been anyway.

tim-s-ccs:
@bjgill Having taken a look, I'm not actually sure what the right way of importing the `DOS6_GO_LIVE_DATE` would be. From my understanding, the `Config` object in `config.py` is loaded in as part of the app. I'm not sure I've done it quite right to by importing the class just to access one variable.

bjgill:
> @bjgill Having taken a look, I'm not actually sure what the right way of importing the `DOS6_GO_LIVE_DATE` would be. From my understanding, the `Config` object in `config.py` is loaded in as part of the app. I'm not sure I've done it quite right to by importing the class just to access one variable.

If you look for `current_app.config` you should be able to see how we do it elsewhere.

tim-s-ccs:
> If you look for `current_app.config` you should be able to see how we do it elsewhere.
> 
> EDIT: it's all rather non-obvious because there's some flask magic going on here. `current_app.config` is how you access the config at run time with all variables updated for the relevant environment and with any variables provided via environment variables loaded.

That is what I tried initially, but I must have messed it up as it didn't work. I've made the change (correctly) now and all looks fine

bjgill (approved):
🎉

